PR #125, opened 2022-01-28
Remove usage of PageWithScrollController

w568w:
We've been always using a `PlatformSubpage` mixing with `PageWithScrollController`, which contains bad practice of introducing local variables into the immutable `StatefulWidget`. 

This PR is to remove usage of `PageWithScrollController`. Also, it moves methods that build `Scaffold` and `AppBar` out of `HomePage`.


singularity-s0:
Build errors detected in this branch by CI system. Try correcting before merging.

singularity-s0 (approved):
All changes reviewed and tested.

PR #126, opened 2022-01-29
Limit the amount of request concurrency

w568w:
At the moment, we are sending requests with different `dio` instances, so they are executed at the same time.

We should not send so many request at one time, since it exerts pressure on the server. This PR contains a `LimitedQueuedInterceptor` to limit the amount of concurrency. 

**However, I am not familiar with concurrent programming. To avoid prospective dead-locking issues, it should be reviewed first.**


singularity-s0 (changes requested):
Minor change requested. No other problems were found during review.

singularity-s0 (approved):

PR #135, opened 2022-02-14
Timetable optimization

w568w:
Allow user to choose semester shown in the timetable. 

Also, I made a bit work on automatically loading predefined start date records of semesters, and applying them to `startDate` when user switches to a semester with predefined record.


singularity-s0 (changes requested):
Start date logic requires improvement.

> 把设置页面里的开学日期设置移动到课程表页面内部

w568w:
> Date does not refresh after changing start date.

I have popped up a tip to notify user to refresh timetable manually.

> I highly doubt whether it is appropriate to use a single provider that passes the SettingsProvider all the way down and notifies all Settings listeners whenever a setting changes.

❓ What do you mean? PTAL [Provider Readme](https://github.com/rrousselGit/provider/blob/master/resources/translations/zh-CN/README.md).

singularity-s0 (approved):

PR #151, opened 2022-04-11
Migrate to JWT Login

### LINKED ISSUE #150: [Feature Request] JWT Authorization
FDUHole is [declared to use JWT Authorization method](https://github.com/OpenTreeHole/treehole_backend) in the backend. Therefore, we have to move on to it as soon as possible.

Some migration work has started in the branch [jwt-login](https://github.com/DanXi-Dev/DanXi/tree/jwt-login). Before this completes, it is hard to begin solving #148 .

**It seems to be a break change, forcing all users to upgrade their application to the next release.** We should keep an eye on potential incompatibility, etc.

w568w:
We are moving to [the new auth service](https://github.com/OpenTreeHole/auth). This PR changes the model definitons and api request methods to adapt to it. 

Also, some fixes to trival issues are contained in the commits.

**Note:** still using the test api endpoint in this PR. We should change it to that of production environment if necessary. Further test is required as well.

It will close #150.


lib/repository/curriculum/curriculum_board_repository.dart
```diff
@@ -17,40 +17,39 @@
...
+  CurriculumBoardRepository._();
 
   static final _instance = CurriculumBoardRepository._();
 
   factory CurriculumBoardRepository.getInstance() => _instance;
 
-  Map<String, String> _tokenHeader(String fduHoleToken) {
-    return {"Authorization": "Token " + fduHoleToken};
+  Map<String, String> _tokenHeader(JWToken fduHoleToken) {
+    return {"Authorization": "Bearer " + fduHoleToken.access!};
```
singularity-s0, line 35: What happens if the server returns a malformed response, i.e. JWT that does not contain `access`? Looks like we are not prepared for that?

lib/repository/opentreehole/opentreehole_repository.dart
```diff
@@ -106,13 +108,14 @@ class OpenTreeHoleRepository extends BaseRepositoryWithDio {
...
       receiveDataWhenStatusError: true,
       validateStatus: (int? status) {
-        if (status == 401) {
-          // If token is invalid, clear the token.
-          SettingsProvider.getInstance().fduholeToken = null;
-        }
         return status != null && status >= 200 && status < 300;
       },
     );
+    dio!.interceptors.add(JWTInterceptor(
```
singularity-s0, line 114: What happens when JWT Refresh token also expires? I did not see code that handles this error, neither here nor in the code of `JWTInterceptor`.

lib/repository/curriculum/curriculum_board_repository.dart
w568w: Indeed, this situation is unexpected but acceptable:
1. If it does not contain necessary field `access`, errors (e.g. `Unauthorized`) will be thrown on the first request, and `JWTInterceptor` will try to require one from the server then. If the interceptor cannot obtain a new token, it will throw an error, allowing user to retry. As for very rare case that `refresh` token is also expired, we will talk about it later.
2. If it does not contain necessary field `refresh`, it will be okay at first if `access` appears, but start reporting an unresolvable `Unauthorized` error after 30 minutes. I am going to deal with it by reinforcing the verifying process at very early stage. (in `OpenTreeHoleRepository.initializeToken()`)

lib/repository/opentreehole/opentreehole_repository.dart
w568w: Likewise,  it starts reporting an unresolvable `Unauthorized` or `Refresh Token Expired` error, until user logs out and signs in again manually. 

I will cope with this situation in the `JWTInterceptor` by extended error type determination process of requests on a new token.

lib/repository/curriculum/curriculum_board_repository.dart
singularity-s0: The use of null check operator complicate debugging. I recommend rejecting malformed JWT in the first place, i.e. throw unauthorized error when malformed response was received.

w568w: >  I recommend rejecting malformed JWT in the first place, i.e. throw error when malformed response was received.

Right. I will check that.

lib/repository/opentreehole/opentreehole_repository.dart
w568w: The new commit c22118cf9f37fa64c27c27c9a544137fdfe9b745 should fix it. PTAL.

lib/repository/curriculum/curriculum_board_repository.dart
w568w: The new commit c22118cf9f37fa64c27c27c9a544137fdfe9b745 should fix it. PTAL.

lib/provider/settings_provider.dart
```diff
@@ -252,11 +252,11 @@ class SettingsProvider with ChangeNotifier {
   set lastPushToken(String? value) =>
       preferences!.setString(KEY_LAST_PUSH_TOKEN, value!);*/
 
-  //Token
+  // Token. If token is invalid, return null.
   JWToken? get fduholeToken {
     if (preferences!.containsKey(KEY_FDUHOLE_TOKEN)) {
       try {
-        return JWToken.fromJson(
+        return JWToken.fromJsonWithVerification(
```
singularity-s0, line 259: If we need verification, why not just make `access` and `refresh` non-nullable. Do we need to write our own methods?

w568w: It is a practice to throw our own error type `BadTokenException` rather than a `NullException`, so we can catch it during logging in or signing up, and distinguish it from other null exception.

singularity-s0 (approved):

PR #162, opened 2022-06-24
Scroll optimize

singularity-s0:
Still need to:

- [x] fix scroll controller


lib/page/platform_subpage.dart
```diff
@@ -167,6 +167,7 @@ abstract class PlatformSubpageState<T extends PlatformSubpage>
     return PrimaryScrollController(
       controller: _buildPrimaryScrollController(context),
       child: PlatformScaffold(
+          backgroundColor: Theme.of(context).scaffoldBackgroundColor,
```
w568w, line 170: What is this line for? Does the `PlatformScaffold` miss a background?

singularity-s0: Yes, on iOS.

lib/page/home_page.dart
```diff
@@ -375,7 +375,7 @@ class HomePageState extends State<HomePage> with WidgetsBindingObserver {
     if (_needSendToWatch &&
         _preferences!.containsKey(SettingsProvider.KEY_FDUHOLE_TOKEN)) {
```
w568w, line 376: Can we eliminate `_preferences` which breaks the uniqueness of `SharedPreferences` held by `SettingsProvider`?

lib/page/platform_subpage.dart
w568w: > Yes, on iOS.

Then is it enough to add the patch *only* here? Has the line covers all `PlatformScaffold`s?

lib/page/home_page.dart
singularity-s0: Yes. Can change this line to use `SettingsProvider`

lib/page/platform_subpage.dart
singularity-s0: I was expecting that `Scaffold`s would automatically use `scaffoldBackgroundColor` from `Theme`, but for some reason it wasn't. This is some kind of workaround solution.

w568w:
Until we figure out a fix to all broken features (i.e. ScrollBar cannot scroll, double tap on title cannot jump to the top), this PR should not be merged.

w568w:
Also PTAL at issue #95. If we uses a paging method like [TieBa](https://tieba.baidu.com/index.html), there should be pretty less problems and bugs.

w568w:
There are some problems preventing this merge:

- [ ] On iOS, after using scroll-to-bottom in `hole_detail`, double tapping on the notification bar (which triggers a scroll in `PrimaryScrollController`) weirdly makes the list scroll to the bottom again, rather than to the top.
- [ ] If entering `hole_detail` with a floor located (e.g. from search results or other's references), the `ScrollBar` on the page looks good but cannot be dragged, throwing an exception indicating that there is no `ScrollPosition` attached to the `PrimaryScrollController` controlled by the scroll bar.
- `scrollable_positioned_list`'s bug: if item's size changes during scrolling by index, list will scroll to an incorrect position. See https://github.com/google/flutter.widgets/issues/330. It seems that the repository is out of maintenance, so we won't expect an official fix for the issue in the short term. (It is NOT a big problem if we drop usage of [keframe](https://github.com/w568w/keframe), because our item widgets' height will not change without it.)

However, removing `keframe` will bring pros and cons. It does have some issues, like slowing down the normal loading speed even under low render pressure, i.e. user is scrolling the list slowly,  but still seeing the following items' transformation from blank to a normal widget. But it really optimizes the frame lag caused by rapid scrolling, giving users a better experience when scrolling.

What do you think towards these problems? @singularity-s0

singularity-s0:
Is it possible to use `Scrollable.ensureVisible()` for locating items instead?

w568w:
> Is it possible to use `Scrollable.ensureVisible()` for locating items instead?

Looks hopeful. You can start a new branch and do some tests on it.

singularity-s0:
> > Is it possible to use `Scrollable.ensureVisible()` for locating items instead?
> 
> Looks hopeful. You can start a new branch and do some tests on it.

Tests complete. You may revert to the previous implementation and merge these branches.

w568w:
The PR and the branch is verified as unworkable. Close it.

w568w:
Can we have a try on <https://github.com/superlistapp/super_sliver_list>? It looks promising.

PR #164, opened 2022-06-30
feat: more color themes

Sunyi618:
more color themes and some tiny modifications

### DIFF
--- a/lib/common/constant.dart
+++ b/lib/common/constant.dart
@@ -376,6 +376,24 @@ class Constant {
     return Colors.red;
   }
 
+  static MaterialColor getMaterialColorFromInt(int color){
+    return MaterialColor(
+      int.parse(color.toString()),
+      <int, Color>{
+        50:  Color(int.parse(color.toString())),
+        100: Color(int.parse(color.toString())),
+        200: Color(int.parse(color.toString())),
+        300: Color(int.parse(color.toString())),
+        400: Color(int.parse(color.toString())),
+        500: Color(int.parse(color.toString())),
+        600: Color(int.parse(color.toString())),
+        700: Color(int.parse(color.toString())),
+        800: Color(int.parse(color.toString())),
+        900: Color(int.parse(color.toString())),
+      },
+    );
+  }
+
   /// A list of Fudan campus.
   ///
   /// It is a copy of [Campus.values] except [Campus.NONE].
--- a/lib/common/constant.dart
+++ b/lib/common/constant.dart
@@ -376,24 +376,6 @@ class Constant {
     return Colors.red;
   }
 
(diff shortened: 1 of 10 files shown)


lib/common/constant.dart
```diff
@@ -376,6 +376,24 @@ class Constant {
...
+        400: Color(int.parse(color.toString())),
+        500: Color(int.parse(color.toString())),
+        600: Color(int.parse(color.toString())),
+        700: Color(int.parse(color.toString())),
+        800: Color(int.parse(color.toString())),
+        900: Color(int.parse(color.toString())),
+      },
+    );
+  }
+
```
w568w, line 396: You needn't hard-code the color palette here. It brings no contrast between widgets and background.

Instead, some libraries are useful to generate color schemes from a single color hex. See: https://github.com/berkanaslan/material-color-generator/blob/main/lib/material_color_generator.dart

lib/widget/dialogs/swatch_picker_dialog.dart
```diff
@@ -32,37 +33,111 @@ class SwatchPickerDialog extends StatefulWidget {
...
+        400: Color(color.value),
+        500: Color(color.value),
+        600: Color(color.value),
+        700: Color(color.value),
+        800: Color(color.value),
+        900: Color(color.value),
+      },
+    );
+  }
+
```
w568w, line 84: Same sugguestion here.

lib/provider/settings_provider.dart
```diff
@@ -460,19 +460,17 @@ class SettingsProvider with ChangeNotifier {
...
     if (preferences!.containsKey(KEY_PRIMARY_SWATCH)) {
-      String color = preferences!.getString(KEY_PRIMARY_SWATCH)!;
-      if (color.trim().isNotEmpty) {
-        return color;
-      }
+      int? color = preferences!.getInt(KEY_PRIMARY_SWATCH);
+      return Color(color!).value;
     }
-    return 'blue';
+    return Colors.blue.value;
```
w568w, line 468: Changing the expected type of stored data may cause type cast exception when user upgrades from old versions. 

Often 2 ways to resolve it:
1. Catch the exception, and set the stored data to the default value, i.e. `Colors.blue.value`.
2. Use another key name in `SettingsProvider`. For example, change `PRIMARY_SWATCH` to `PRIMARY_SWATCH_V2`.

Sunyi618:
Done

w568w (approved):
LGTM. 

However `SettingsProvider.primarySwatch_V2` does not correspond to [Dart code style](https://dart.cn/guides/language/effective-dart/style#do-name-other-identifiers-using-lowercamelcase). Consider renaming it as original `primarySwatch`, which will cause no problem.

PR #167, opened 2022-10-04
Feat: adding courses manually

Sunyi618:
Finally it works. 🥰


lib/page/subpage_timetable.dart
```diff
@@ -113,21 +141,45 @@ class TimetableSubPageState extends PlatformSubpageState<TimetableSubPage> {
...
       GlobalKey<RefreshIndicatorState>();
 
   void pageRefresh() {
     refreshSelf();
   }
 
+  List<Course> getNewCourseList() {
+    SharedPreferences preferences = SettingsProvider.getInstance().preferences!;
+    if (!preferences.containsKey(Constant.KEY_MANUALLY_ADDED_COURSE)) {
+      preferences.setString(Constant.KEY_MANUALLY_ADDED_COURSE, "[]");
```
w568w, line 159: Default preference value should be set in the corresponding getter method of `SettingsProvider`, not here.

lib/page/subpage_timetable.dart
```diff
@@ -154,22 +206,34 @@ class TimetableSubPageState extends PlatformSubpageState<TimetableSubPage> {
...
         }
       }
       forceLoadFromRemote = false;
     } else {
       _contentFuture = LazyFuture.pack(Future<TimeTable?>.error(
-          NotLoginError(S.of(context).not_fudan_student)));
+          NotLoginError(S.of(context).not_fudan_student)))
+        ..then(
+            (value) => TimeTable.mergeManuallyAddedCourses(value, newCourses))
+        ..then((value) => temp = value);
```
w568w, line 236: If you want:

```dart
if (...) {
    A();
    commonLogics();
} else if (...) {
    B();
    commonLogics();
} else {
    C();
    commonLogics();
}
```
Then why not use:
```dart
if (...) {
    A();
} else if (...) {
    B();
} else {
    C();
}
commonLogics();
```

lib/page/subpage_timetable.dart
```diff
@@ -241,7 +306,68 @@ class TimetableSubPageState extends PlatformSubpageState<TimetableSubPage> {
...
+                    builder: (_) => ManuallyAddCourseDialog(
+                      courseAvailableList,
+                    ),
+                  )?.then<List<Course>>((course) {
+                    List<Course> oldCourseList = getNewCourseList();
+                    oldCourseList.add(course!);
+                    return oldCourseList;
+                  }))!, (cachedValue) {
+            List<String>? courseListString =
+                cachedValue?.substring(1, cachedValue.length - 1).split(", ");
```
w568w, line 325: Same here. If you would like to serialize/deserialize something (i.e. storing and loading them as strings), ensure that there is no ready-made method for you to use.

It is good to play with strings in programming course homework, but I would rather not suggest them in a project.

Issue #235, opened 2023-04-24
[Feature Request] 搜索自杀关键词时弹出救援热线


KYLN24:
搜索自杀关键词时弹出救援热线

Boreas618:
In commit 52f547fe74e7aab2e2825e600c7a7c042f705186, we introduced **DanXi Care** on the search page.

Some **care words** are suggested to detect potential suicide-like incidents. You can modify this list in the source code's assets or on https://github.com/DanXi-Dev/DanXi-Backend/blob/main/all.json.

Future directions include:

- More professional care messages (the current message is generated by GPT-4).
- Detection of care needs when sending posts.

<img width="353" alt="Screenshot 2023-07-19 at 4 08 17 PM" src="https://github.com/DanXi-Dev/DanXi/assets/98612013/bc517e20-8b29-49cc-9483-ff0220a5ae3c">

w568w:
这个实现完了吗？如果实现完了请 close。

kooWZ:
> 这个实现完了吗？如果实现完了请 close。

文案还需要调整
之前群里讨论的文案还没放进去

PR #262, opened 2023-07-27
feat: show DanXi Care dialog when attempting to send posts with care words

kooWZ:
Issue #235 
The first time a user attempts to send posts with care words, DanXi Care dialog box will pop up, and the message will not be sent.
Only when clicking again will the post be actually sent.

PS. What's the necessity to use `WidgetsBinding.instance.addPostFrameCallback` here and in commit 52f547f ?

[[lib/page/opentreehole/hole_search.dart](https://github.com/DanXi-Dev/DanXi/blob/52f547fe74e7aab2e2825e600c7a7c042f705186/lib/page/opentreehole/hole_search.dart#L176)](https://github.com/DanXi-Dev/DanXi/blob/52f547fe74e7aab2e2825e600c7a7c042f705186/lib/page/opentreehole/hole_search.dart#L176-L183)

### DIFF
--- a/lib/page/opentreehole/hole_editor.dart
+++ b/lib/page/opentreehole/hole_editor.dart
@@ -28,11 +28,13 @@ import 'package:dan_xi/provider/fduhole_provider.dart';
 import 'package:dan_xi/provider/settings_provider.dart';
 import 'package:dan_xi/repository/opentreehole/opentreehole_repository.dart';
 import 'package:dan_xi/util/browser_util.dart';
+import 'package:dan_xi/util/danxi_care.dart';
 import 'package:dan_xi/util/master_detail_view.dart';
 import 'package:dan_xi/util/noticing.dart';
 import 'package:dan_xi/util/opentreehole/editor_object.dart';
 import 'package:dan_xi/util/platform_universal.dart';
 import 'package:dan_xi/util/public_extension_methods.dart';
+import 'package:dan_xi/widget/dialogs/care_dialog.dart';
 import 'package:dan_xi/widget/libraries/error_page_widget.dart';
 import 'package:dan_xi/widget/libraries/image_picker_proxy.dart';
 import 'package:dan_xi/widget/libraries/linkify_x.dart';
@@ -605,6 +607,7 @@ class BBSEditorPageState extends State<BBSEditorPage> {
 
   bool _isFullscreen = false;
   bool? _supportTags;
+  bool _confirmCareWords = false;
 
   String? _tip;
   late EditorObject _object;
@@ -680,7 +683,23 @@ class BBSEditorPageState extends State<BBSEditorPage> {
               icon: PlatformX.isMaterial(context)
                   ? const Icon(Icons.send)
                   : const Icon(CupertinoIcons.paperplane),
-              onPressed: _canSend ? () => _sendDocument(_object) : null),
+              onPressed: _canSend
+                  ? () async {
+                      bool isCareWordsDetected =
+                          await detectCareWords(_controller.text);
+                      WidgetsBinding.instance.addPostFrameCallback((_) async {
+                        if (isCareWordsDetected && _confirmCareWords == false) {
+                          await showPlatformDialog(
+                              context: context,
+                              builder: (_) => const CareDialog());
+                          _confirmCareWords = true;
+                          return;
+                        }
+                        _sendDocument(_object);
+                      });
+                    }
+                  : null,
+          ),
         ],
       ),
       body: SafeArea(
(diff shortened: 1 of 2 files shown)


w568w:
Look almost good to me.

> What's the necessity to use `WidgetsBinding.instance.addPostFrameCallback` here and in commit https://github.com/DanXi-Dev/DanXi/commit/52f547fe74e7aab2e2825e600c7a7c042f705186 ?

No idea and I do not suppose it is essential. Maybe ask @Boreas618.

Boreas618:
> What's the necessity to use WidgetsBinding.instance.addPostFrameCallback here and in commit https://github.com/DanXi-Dev/DanXi/commit/52f547fe74e7aab2e2825e600c7a7c042f705186 ?

To address the warning about the **asynchronous build context**. This is not a recommended practice as it can be confusing to understand. Perhaps I can refactor this section using the instructions provided [here](https://dart.dev/tools/linter-rules/use_build_context_synchronously).

Boreas618:
In my opinion, we can implement the detection of care words using the `PostInterceptor`. This would result in more consistent source code. Please refer to the implementation of `_kStopWordInterceptor` for an example.

w568w:
> > What's the necessity to use WidgetsBinding.instance.addPostFrameCallback here and in commit [52f547f](https://github.com/DanXi-Dev/DanXi/commit/52f547fe74e7aab2e2825e600c7a7c042f705186) ?
> 
> To address the warning about the **asynchronous build context**. This is not a recommended practice as it can be confusing to understand. Perhaps I can refactor this section using the instructions provided [here](https://dart.dev/tools/linter-rules/use_build_context_synchronously).

Nope... I do not think it will address the warning. It just *bypasses* the warning in a way that the linter is not smart enough to recognize, and indeed makes the thing worse: 

Linter will complain when using a `context` after an asynchronous gap, because we cannot know whether the `context` is still effective after the gap. Look at the code below:

```dart
// We are in a normal button's onTap() callback.
final text = _controller.text;
await aVeryLongOperation(text); // this operation takes very long, up to 1 hour!
Noticing.showNotice(context, "Hey, We Are Done"); // use the context across the "asynchronous gap" above
```

Say that a user clicks the button, and before `aVeryLongOperation()` gets done, he exits the page containing this button, so the button gets "destroyed". 

Now what about the running `onTap`? Will it be canceled? No. Actually it executes until the last line. When `await aVeryLongOperation` finally finishes, `context` is used in calling `Noticing.showNotice` then. However, since the button has been destroyed a long time ago, of course the `context` is invalid now (or specifically, `context.mounted == false`), and `Noticing.showNotice` will eventually get crashed. That's why we should check whether `context.mounted` after awaiting an asynchronous function (i.e. "asynchronous gap").

So if you put `Noticing.showNotice(context, "Hey, We Are Done")` in a `addPostFrameCallback`, you are at one step further to postponing the usage of `context`, making it more likely to have become unusable when reaching for it; thus, let the thing be worse.

Boreas618:
Got it

> > > What's the necessity to use WidgetsBinding.instance.addPostFrameCallback here and in commit [52f547f](https://github.com/DanXi-Dev/DanXi/commit/52f547fe74e7aab2e2825e600c7a7c042f705186) ?
> > 
> > 
> > To address the warning about the **asynchronous build context**. This is not a recommended practice as it can be confusing to understand. Perhaps I can refactor this section using the instructions provided [here](https://dart.dev/tools/linter-rules/use_build_context_synchronously).
> 
> Nope... I do not think it will address the warning. It just _bypasses_ the warning in a way that the linter is not smart enough to recognize, and indeed makes the thing worse:
> 
> The reason why linter will complain when using a `context` after an asynchronous gap is that, we cannot know whether the `context` is still effective after the gap. Look at the code below:
> 
> ```dart
> // We are in a normal button's onTap() callback.
> final text = _controller.text;
> await aVeryLongOperation(text); // this operation takes very long, up to 1 hour!
> Noticing.showNotice(context, "Hey, We Are Done"); // use the context across the "asynchronous gap" above
> ```
> 
> Say that a user clicks the button, and before `aVeryLongOperation()` gets done, he exits the page containing this button, so the button gets "destroyed".
> 
> Now what about the running `onTap`? Will it be canceled? No. Actually it executes until the last line. When `await aVeryLongOperation` finally finishes, `context` is used in calling `Noticing.showNotice` then. However, since the button has been destroyed a long time ago, of course the `context` is invalid now (or specifically, `context.mounted == false`), and `Noticing.showNotice` will eventually get crashed. That's why we should check whether `context.mounted` after awaiting an asynchronous function (i.e. "asynchronous gap").
> 
> So if you put `Noticing.showNotice(context, "Hey, We Are Done")` in a `addPostFrameCallback`, you are at one step further to postponing the usage of `context`, making it more likely to have become unusable when reaching for it; thus, let the thing be worse.

kooWZ:
> In my opinion, we can implement the detection of care words using the `PostInterceptor`. This would result in more consistent source code. Please refer to the implementation of `_kStopWordInterceptor` for an example.

I think it would be complicated to achieve the effect of sending posts only on the second click using `PostInterceptor`.

Boreas618:
LGTM

PR #278, opened 2023-08-09
fix overflow in SearchHistory

### LINKED ISSUE #277: [BUG] 历史搜索页面overflow
当历史搜索数量超出一屏时，
https://github.com/DanXi-Dev/DanXi/blob/97c48f6f90a74be6b22559a7d631d66894a6b959/lib/page/opentreehole/hole_search.dart#L133
报overflow，同时内部的ListView不能滚动

kooWZ:
The column of search history widget is nested in the main column, leaving the size of search history column uncertain.
This causes overflow when wrapping `ListView` in `Flexible`, or exception when wrapping it in `Expanded`.

After enclosing `Column` in `Expanded`, both `Flexible` and `Expanded` works well with `ListView`.
Closes #277

### DIFF
--- a/lib/page/opentreehole/hole_search.dart
+++ b/lib/page/opentreehole/hole_search.dart
@@ -180,8 +180,8 @@ Widget searchByText(BuildContext context, String searchKeyword) {
               await showPlatformDialog(
                   context: context, builder: (_) => const CareDialog());
             }
-              smartNavigatorPush(_globalKey.currentContext!, "/bbs/postDetail",
-                  arguments: {"searchKeyword": searchKeyword});
+            smartNavigatorPush(_globalKey.currentContext!, "/bbs/postDetail",
+                arguments: {"searchKeyword": searchKeyword});
           },
         );
       });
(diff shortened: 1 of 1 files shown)


lib/page/opentreehole/hole_search.dart
```diff
@@ -58,7 +58,10 @@ class _OTSearchPageState extends State<OTSearchPage> {
   /// The text user inputs.
   final TextEditingController _searchFieldController = TextEditingController();
 
-  Widget _buildSearchHistory(BuildContext context) => Column(
+  /// Use Expanded to avoid the uncertainty of column size,
+  /// thus solving Flexible overflow
+  Widget _buildSearchHistory(BuildContext context) => Expanded(
```
w568w, line 63: It is not recommended to return a `Expanded` Widget, since it is only valid inside a `Flex` Widget (e.g. `Column`, `Row`, `Flex`). Without proper comments, it might result in coupled modifications when the parent layout is changed later on.

You should, instead, put the wrapping widget in where the parent `Flex` is declared, e.g.

```dart
Column(
  children: [
    // ...
    Expanded(child: _buildSearchHistory(context)),
    // ...
  ]
)
```

w568w (changes requested):

lib/page/opentreehole/hole_search.dart
kooWZ: Got it
Moved `Expanded`  inside main `Column`

w568w (approved):
Tested. LGTM.

PR #295, opened 2023-09-30
fix: wrong auth url & unable to login to fduhole

kooWZ:


lib/repository/opentreehole/opentreehole_repository.dart
```diff
@@ -147,10 +147,18 @@ class OpenTreeHoleRepository extends BaseRepositoryWithDio {
...
-        "$_BASE_AUTH_URL/verify/email",
-        queryParameters: {"email": email, "check": true});
-    return response.data!["registered"];
+    try {
+      final Response<Map<String, dynamic>> response = await secureDio.get(
+          "$_BASE_AUTH_URL/verify/email",
+          queryParameters: {"email": email, "check": true});
+      return response.data!["registered"];
+    } catch (e) {
+      if (e is DioError && e.response!.data["registered"] != null) {
```
w568w, line 156: I guess you want to catch a non-2xx HTTP response code here? 

If so, just use [`validateStatus`](https://pub-web.flutter-io.cn/packages/dio#request-options) to override how Dio validates error from HTTP status codes.

kooWZ: Got it.
Is this ideal?
```dart
  Future<bool?> checkRegisterStatus(String email) async {
    final Response<Map<String, dynamic>> response = await secureDio.get(
        "$_BASE_AUTH_URL/verify/email",
        queryParameters: {"email": email, "check": true},
        options: Options(
            validateStatus: (status) => status != null && status == 400));
    if (response.data!["registered"] != null) {
      return response.data!["registered"];
    } else {
      throw (response.data!["message"]);
    }
  }
```

w568w: > Got it. Is this ideal?

I would recommend replace `status == 400` with `status <= 400`, otherwise you are viewing success responses (e.g. `200 OK`) as errors too.

kooWZ: > > Got it. Is this ideal?
> 
> I would recommend replace `status == 400` with `status <= 400`, otherwise you are viewing success responses (e.g. `200 OK`) as errors too.

OK. But as per backend, this request will only return 400.

w568w (approved):

PR #303, opened 2023-10-18
New admin punishment page

HydrogenC:
WIP, shall close #261.

PR #305, opened 2023-10-21
Fix layout spacing problems

w568w:
Cherry picked from #303 for 1.4.0.

### DIFF
--- a/lib/page/subpage_danke.dart
+++ b/lib/page/subpage_danke.dart
@@ -37,6 +37,7 @@ import 'package:dan_xi/widget/libraries/future_widget.dart';
 import 'package:dan_xi/widget/libraries/sized_by_child_builder.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter_platform_widgets/flutter_platform_widgets.dart';
 import 'package:provider/provider.dart';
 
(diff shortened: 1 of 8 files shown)


singularity-s0:
Exact what problem does this PR fix?

w568w:
> Exact what problem does this PR fix?

Too big/small paddings, layout overflows... Something like these.

lib/page/subpage_danke.dart
```diff
@@ -75,94 +79,105 @@ class DankeSubPageState extends PlatformSubpageState<DankeSubPage> {
...
+          key: indicatorKey,
+          edgeOffset: MediaQuery.of(context).padding.top,
+          color: Theme.of(context).colorScheme.secondary,
+          backgroundColor: Theme.of(context).dialogBackgroundColor,
+          onRefresh: () async {
+            HapticFeedback.mediumImpact();
+            await _loadRandomReview(forceRefetch: true);
+            setState(() {});
+          },
+          child: SizedByChildBuilder(
```
w568w, line 92: Why use `SizedByChildBuilder` here, as you do not seem to use the measured `size` later at all?

cc @HydrogenC

lib/page/subpage_danke.dart
```diff
@@ -75,94 +79,105 @@ class DankeSubPageState extends PlatformSubpageState<DankeSubPage> {
...
-                    stackTrace: snapshot.stackTrace,
-                    onTap: () => setState(() {}));
-              }
-            },
-            loadingBuilder: Center(
-              child: PlatformCircularProgressIndicator(),
-            ))
-        // A maximized result list
-        : Expanded(child: CourseListWidget(searchKeyword: searchText));
+    return Expanded(
```
w568w, line 139: Again, do not use `Expanded` as your root widget in a method or a custom widget's `build()`. 

cc @HydrogenC

w568w (commented):
Minor changes.

lib/page/subpage_danke.dart
HydrogenC: The top padding used to be height-dependent of search bar, but I forgot to remove this after the top padding was cancelled.

lib/page/subpage_danke.dart
HydrogenC: My bad.

PR #324, opened 2023-12-14
change: repick icon for adding course review

HydrogenC:
Change the icon for adding course review to `add` in favor of the feedback from the user group chat.


w568w:
LGTM!

PR #325, opened 2023-12-21
Fix #226 and upgrade dependencies (Fixed)

w568w:
This PR fixes #226 and upgrade dependencies.

This PR is a subsequent version of #316, containing the changes:

- Downgrade `flutter_secure_storage` and `encrypt_shared_preferences` for now, before we found why it was broken;
- Update to Flutter 3.16.3.

I also applied `dart fix` to adapt to the recently introduced `super.xxx` syntax.

PR #327, opened 2024-01-01
Fix #326

### LINKED ISSUE #326: [BUG] [Help wanted] 安卓设备上无法修改课评
**描述 Bug**
电脑端上课评修改功能工作正常，提交后显示请求成功，课评也成功更改。  
安卓端提交修改后app不发送请求。  
来自社区反馈，机型未知。但在本人的小米K60上成功复现此问题，尚不知该问题是否影响所有机型。

**复现步骤**
复现该 Bug 的步骤：
1. 找到自己发送的课评
2. 修改并提交

**截图**
无

**系统信息**
 - 设备型号：Xiaomi K60
 - 系统版本：Android 13
 - 旦夕版本（见「设置-关于本应用」）：1.4.1

w568w:
As the title says.

See #326 for more details. This pull request also removes some possibly poorly flavored codes.

Fixes #326.

### DIFF
--- a/lib/page/danke/course_review_editor.dart
+++ b/lib/page/danke/course_review_editor.dart
@@ -581,7 +581,7 @@ class CourseRatingWidgetState extends State<CourseRatingWidget> {
             mainAxisSize: MainAxisSize.min,
             mainAxisAlignment: MainAxisAlignment.spaceBetween,
             children: [
-              SizedBox(width: 80, child: Text(widget.label)),
+              SizedBox(width: 88, child: Text(widget.label)),
               Row(
                 mainAxisSize: MainAxisSize.min,
                 children: List.generate(
--- a/lib/page/danke/course_review_editor.dart
+++ b/lib/page/danke/course_review_editor.dart
@@ -579,8 +579,9 @@ class CourseRatingWidgetState extends State<CourseRatingWidget> {
         padding: const EdgeInsets.symmetric(vertical: 4),
         child: Row(
             mainAxisSize: MainAxisSize.min,
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
             children: [
-              SizedBox(width: 88, child: Text(widget.label)),
+              SizedBox(width: 80, child: Text(widget.label)),
               Row(
                 mainAxisSize: MainAxisSize.min,
                 children: List.generate(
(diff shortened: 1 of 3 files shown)


lib/page/danke/course_review_editor.dart
```diff
@@ -579,8 +579,9 @@ class CourseRatingWidgetState extends State<CourseRatingWidget> {
         padding: const EdgeInsets.symmetric(vertical: 4),
         child: Row(
             mainAxisSize: MainAxisSize.min,
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
             children: [
-              SizedBox(width: 88, child: Text(widget.label)),
+              SizedBox(width: 80, child: Text(widget.label)),
```
HydrogenC, line 584: 80 works better for Chinese, but I found that it isn't enough for Japanese or English, so I set it to 88. I believe the value should remain 88 before any better solution is found.

HydrogenC (changes requested):
Thank you a lot for refractoring and removing those very old code that should have been deleted. It looks perfect now!

lib/page/danke/course_review_editor.dart
w568w: > 80 ... isn't enough for Japanese or English, so set it to 88

OK, then I will revert it. 

`88px` caused an 8.2px layout overflow on my Android device, so I tuned it a bit. 

IMHO, the best solution is that a hard-coded dimension should not be used here. I think that you actually want to **align text labels and star selectors**, and `Column` + `Row` is not the solution. Please take a look at [flutter_layout_grid](https://pub.dev/packages/flutter_layout_grid).

HydrogenC: It doesn't sound worth adding a new dependency for this tiny feature. I will try to implement this by myself with existing widget. It won't be too difficult, I guess.

w568w: > It doesn't sound worth adding a new dependency for this tiny feature

So we already depend on [flutter_layout_grid](https://pub.dev/packages/flutter_layout_grid) for timetable layout.

> this doesn't actually solve the overflow under other languages

Yes. Check out our another dependency, [auto_size_text](https://pub.dev/packages/auto_size_text).

w568w:
After you approve it, we will merge it.

HydrogenC (approved):

w568w:
Now wait for #325 to merge. ;)

PR #328, opened 2024-01-05
Adopt layout grid for column alignment

HydrogenC:
Adoptions: 
- Rating widget in course review editor
- Developer list


HydrogenC:
Please review @w568w

lib/page/danke/course_review_editor.dart
```diff
@@ -418,36 +438,68 @@ class CourseReviewEditorWidgetState extends State<CourseReviewEditorWidget> {
...
+            LayoutGrid(
+              columnSizes: [auto, 1.fr, auto],
+              rowSizes: const [auto, auto, auto, auto],
+              children: [
+                Center(child: Text(S.of(context).curriculum_ratings_overall)),
+                RatingStarWidget(
+                  initialRating: review.grade.overall,
+                  onRate: (e) {
+                    review.grade = review.grade.withFields(overall: e);
+                    _ratingChangedStream.add(RatingChangedEvent(review.grade));
```
w568w, line 450: I guess that you use `StreamController` to notify other widgets about the changes of rating. It works but does not represent our best practice: using `Provider` to maintain the state consistency among different widgets. 

You may have to change your view on how one widget affects each other:

**Now:**

```
Page------>Widget1
   |          |
   |          |
   |          |Stream.add()
   |          |
   |          |
   |          v
   +------>Widget2
```

**Better:**

```
       Page------>Widget1---+
(With a Provider)           |
          |                 |    +----------------------+
          |        Listening|    |      Provider        |
          |                 |    |                      |
          |                 +----+---> Rating: 1        |
          |                      |        ^             |
          +------>Widget2--------+--------+             |
                                 |Listening             |
                                 |                      |
                                 |                      |
                                 +----------------------+
```

When `Widget1` needs to change states, instead of sending an `Event` to everyone, it had better change states in `Provider` and anyone else listening to the same state will be notified (i.e. re-`build()`) at once. This is called **Reactive Programming**.

Unfortunately `Provider`'s **official documentation is really poorly written and focuses on troubleshooting and listing everything, does not introduce its usage and design at all**. Flutter's documentation is not much better. You can check out [this article](https://juejin.cn/post/7266737898668589091) for a quick start.

w568w (changes requested):
Sorry for being really busy these days.

lib/page/danke/course_review_editor.dart
HydrogenC: Okay, I understand. I'm sorry for sticking too close to the vanilla style and not utilizing some popular and nice packages.

HydrogenC:
I took advantage of provider in my new code. It seems to work fine, but I'm not sure whether I should add anything to didChangeDependencies or didUpdateWidget.

lib/page/danke/course_review_editor.dart
```diff
@@ -348,173 +338,193 @@ class CourseReviewEditorWidgetState extends State<CourseReviewEditorWidget> {
...
-                              initialSelection: selectedCourse,
-                              items: value == "*"
-                                  ? []
-                                  : widget.courseGroup.courseList!
-                                      .filter((e) => e.teachers == value),
-                              hintText: S.of(context).curriculum_select_time,
-                              labelText: S.of(context).course_schedule,
+    return ChangeNotifierProvider(
+        create: (_) => review.grade.clone(),
+        builder: (ctx, _) => SingleChildScrollView(
```
w568w, line 343: You do not really need to pass the `builder` argument here because you are already using `Consumer` to listen to the `CourseGrade` everywhere and your widget tree is ***stateless*** now (by Flutter's definition. i.e. the methods and callbacks in this widget tree do not call `setState()` explicitly at all, and the tree no longer relies on Flutter's state management: yes, you are instead using `Provider`'s state management!). 

For a stateless widget, you should pass it to `child` argument, rather than `builder`. The latter will be usually only used for a convenient way where you do not want to use `Consumer` widget. For example:

```dart
builder: (ctx, child) {
  final grade = Provider.of<CourseGrade>(ctx, listen: true);
  return Text(grade.xxx);
}
``` 

is **_equivalent_** to:

```dart
child: Consumer<CourseGrade>(
  builder: (ctx, grade, child) => Text(grade.xxx))
```

w568w (changes requested):
> I'm not sure whether I should add anything to didChangeDependencies or didUpdateWidget.

You should implement `didUpdateWidget` if you save some widget arguments in your `State` class (e.g. `late CourseReviewEditorText review;`).

`didUpdateWidget` is called when a widget argument has been changed; `didChangeDependencies` is called when your widget dependency has been changed ("widget dependency" is about [`InheritedWidget`](https://api.flutter.dev/flutter/widgets/InheritedWidget-class.html) mechanism, a Flutter mechanism that `Provider` relies on internally.) and after `initState()`.

Documentation says `didChangeDependencies` is "**rarely** overrided  because the framework always calls [`build`](https://api.flutter.dev/flutter/widgets/State/build.html) after a dependency changes. Some subclasses do override this method because they need to do some expensive work (e.g., network fetches) when their dependencies change, and that work would be too expensive to do for every build." 

In your case, just using `build()` is good enough.

HydrogenC:
Sorry for being slow.

lib/repository/opentreehole/opentreehole_repository.dart
```diff
@@ -146,6 +148,16 @@ class OpenTreeHoleRepository extends BaseRepositoryWithDio {
     }
   }
 
+  /// A full initialization of the user data.
+  ///
+  /// It loads the token and user info which are shared across treehole and danke.
+  ///
+  /// It is used to provide user data for sections apart from treehole (specifically danke for now).
+  Future<void> initializeUser() async {
```
w568w, line 156: I am wondering if it is necessary to provide 3 different initializers. Well, I have no better proposal for now.

w568w (approved):
LGTM. Tested on Linux Desktop.

lib/repository/opentreehole/opentreehole_repository.dart
HydrogenC: It requires a more thorough refractor to decouple the Danxi account system from treehole. To make a quick fix for this annoying problem, I can only do this.

PR #330, opened 2024-01-12
fix: show grade for 初选免听/重修免听 courses

kooWZ:
Grade not shown caused by inconsistency of name between two data sources.
Matching course ID may be a better approach.

image


w568w (approved):
LGTM. I never considered this case before...

PR #334, opened 2024-03-08
fix: scroll optimize

ivanfei-1:
The super_sliver_list package seems to fix this long existing problem mentioned in pr #162. Since the branch is long gone, I am opening a new one.

I have tested the performance on the material-2 branch, where the improvement is significant.

Before merging this branch, you should test on your own device with different version of flutter.

### DIFF
--- a/lib/widget/libraries/paged_listview.dart
+++ b/lib/widget/libraries/paged_listview.dart
@@ -25,6 +25,7 @@ import 'package:dan_xi/widget/libraries/future_widget.dart';
 import 'package:dan_xi/widget/libraries/state_key.dart';
 import 'package:dan_xi/widget/libraries/with_scrollbar.dart';
 import 'package:flutter/material.dart';
+import 'package:super_sliver_list/super_sliver_list.dart';
 
 const kDuration = Duration(milliseconds: 500);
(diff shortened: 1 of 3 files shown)


w568w:
Some problems noticed:

1. On Android, in a very deep post (e.g. #60678), if "jump to the bottom" is tapped and finding any floor A which refers to another floor B in the same post, trying jumping to floor B by tapping "Locate" in A's reference will always scroll the list to the top, rather than scroll to B.

w568w:
\*sigh\* 

I think we need to overhaul the entire `PagedListView` (and `StateKey`), because there are some implementations that don't follow the Flutter state specification at all, and there are a lot of hacks. They're very likely to cause mysterious bugs.

w568w:
Update: the 2nd problem seems to be introduced in 2d260ce998d47df62d6ca1b35564186c4185a231, where the `_hole.floors!.prefetch` is reset to only first 10 items during each build.

Anyway, I will rewrite this part. It feels really awkward now.

ivanfei-1:
LGTM. Basic featured have been tested, and no bug is found so far. I think we can merge this branch before thorough testing.

PR #336, opened 2024-03-13
Fix inconsistent results caused by course review editor

HydrogenC:
Community-risen issue from the Danxi group chat.


w568w (approved):
LGTM.

PR #349, opened 2024-04-22
fix: empty classroom not showing correctly

kooWZ:
The website structure may be updated and there is no need to ignore first `<td>` now, since it will not be recognized by regex.


w568w (approved):
Minor changes. LGTM.

PR #352, opened 2024-04-29
[Fix] Timetable unable to load for postgraduates

PramSin:
There're two changes:
1. Do not use the `atob` function, only use the supported JavaScript functions in flutter_js to perform `base64` decoding.
2. If `course.roomname` is `null`, then assign it as a blankspace `' '`

Tested on android already with only my account.
The preview is below
image


w568w (approved):
LGTM!

PR #374, opened 2024-07-01
fix: #370

w568w:
- Linux 打开黑白屏：已在 PR 中修复；
- Windows 无法启动：和报告用户跟进，确定是数据问题，删除 `C:/users/用户名/AppData/Roaming/io.github.w568w/dan_xi/` 后恢复正常。


singularity-s0 (approved):
This is probably due to some kind of optimization where UI elements behind an opaque picture no longer needs to be rendered. But Flutter probably messed up somewhere. This shouldn't be our problem.

PR #377, opened 2024-07-01
UI tweaks for forum (1st Pass)

### LINKED ISSUE #365: [Feature Request] 茶楼UI微调
**你的功能需求和某个bug有关吗？**
否

**你想要什么样的功能？**
茶楼（原树洞）的UI需要微调，没有明确标准，和原来的UI能够明显区分即可。

HydrogenC:
Tweaks the UI towards the style of the swift version, resolves #365. Might require some additional polishing if any artifact is found about the layout.  
Showcase:  
e7716538f82d397aa7109a34952fcdf

6c214352b0d9ebd77c45094e763d428


HydrogenC:
Sin all operations in the popup menu prompted by the `...` button also exists in the modal sheet. So I removed the popup menu and redirected the `...` button to the modal sheet.

w568w:
> Sin all operations in the popup menu prompted by the `...` button also exists in the modal sheet. It's a sort of duplication, so I removed the popup menu and redirected the `...` button to the modal sheet.

That is what I've wanted to do for a long time!

lib/widget/forum/forum_widgets.dart
```diff
@@ -325,7 +249,42 @@ class OTHoleWidget extends StatelessWidget {
           } finally {
             dialog.dismiss(showAnim: false);
           }
-        });
+        },
+        child: IntrinsicHeight(child: Padding(
```
w568w, line 253: What is this `IntrinsicHeight` for?

w568w (approved):
Mostly LGTM. Wonderful work!

lib/widget/forum/forum_widgets.dart
HydrogenC: I don't want to hardcode the height of the ColoredBox, so I used `CrossAxisAlignment.stretch` to keep it the same height as the actual content. But doing this will cause a unbounded height, according to StackOverflow, I have to wrap the row in an `IntrinsicHeight` to make `CrossAxisAlignment.stretch` to work properly.

HydrogenC:
Also resolves #371.

w568w:
You do not need to fix 064d6fdeff9c7c5c479ceaf7128ac8752f335943!

I have already done in #382 .

lib/widget/forum/forum_widgets.dart
w568w: > @w568w I don't want to hardcode the height of the ColoredBox, so I used `CrossAxisAlignment.stretch` to keep it the same height as the actual content. But doing this will cause a unbounded height, according to StackOverflow, I have to wrap the row in an `IntrinsicHeight` to make `CrossAxisAlignment.stretch` to work properly. Is there a better way to do this?

Yeah you are doing it properly, but with an extra layouting, which is quite expensive. A more efficient way is using [boxy](https://pub.dev/packages/boxy) for conditional layout constraints. However I believe the current way is good enough ;)

w568w (approved):

PR #382, opened 2024-07-02
add http proxy support

### LINKED ISSUE #369: [Feature Request] 支持应用内设置代理
由于目前诸多服务需要校内环境访问，且 Flutter 自身似乎不会读取环境变量中的代理设置，使用 [docker-easyconnect](https://github.com/docker-easyconnect/docker-easyconnect/) 等代理技术的用户将无法很好地在桌面端使用。

应当在设置中增加代理选项，允许用户配置自定义代理。

下面是一个例子：

```dart
(dio.httpClientAdapter as DefaultHttpClientAdapter).onHttpClientCreate =
        (client) {
      client.findProxy = (uri) => "PROXY 127.0.0.1:8888;";
      client.badCertificateCallback =
          (X509Certificate cert, String host, int port) => true;
    };

```

目前，仅需考虑 HTTP 代理即可。

w568w:
This PR will:

1. support setting HTTP proxy (close #369);
2. add backward compatibility for previous `danxi_` sticker prefix;
3. remove some useless translations.

**Note**: after this PR, some code behaviors change:

1. For any `new Dio(...)`, use `DioUtils.newDioWithProxy(...)` instead;
2. For any `DefaultCacheManager()`, use `DefaultCacheManagerWithProxy()` instead;
3. For any `CachedNetworkImage`, add `cacheManager: DefaultCacheManagerWithProxy()` argument.


singularity-s0:
Why add a separate proxy setting instead of respecting the system configuration?

w568w:
> Why add a separate proxy setting instead of respecting the system configuration?

1. **There is never a "standard way" to detect proxy on each platform.** 
    - For Windows, it is put at `HKCU\Software\Microsoft\Windows\CurrentVersion\Internet Settings`; 
    - For Linux, `all|socks|http(s)_proxy` envs have never been standardized, each desktop environment develops its own properties for proxy setting, and the more modern `libproxy` is still on its way to be widely accepted;
    - For Android and iOS, [things get even harder](https://github.com/vicdotdevelop/native_flutter_proxy/blob/master/ios/Classes/SwiftFlutterProxyPlugin.swift): they are scattered in JVM environments and official APIs. 
2. Due to the API limit, **we can only support plaintext HTTP proxy** at the moment. On some systems, the "system proxy" can be of other proxy type. We cannot respect that.
3. Campus VPN is such a specific requirement that: **for most users, only DanXi requires it**. Why does a user have to set his system proxy to EasyConnect, just to make DanXi happy?

singularity-s0 (approved):

PR #388, opened 2024-07-03
Show floor bottom bar even if deleted

### LINKED ISSUE #387: [Feature Request][Internal] 已删除的层显示id和时间
**你的功能需求和某个bug有关吗？**
`[否]`

**你想要什么样的功能？**
已删除的层同样显示id和时间，当出现用户连续删除多个层时，方便管理员定位

HydrogenC:
Supposed to resolve #387, but untested due to lack of PC devive.

### DIFF
--- a/lib/widget/forum/forum_widgets.dart
+++ b/lib/widget/forum/forum_widgets.dart
@@ -452,7 +452,7 @@ class OTFloorWidget extends StatelessWidget {
                         onLinkTap,
                         onTapImage ?? defaultOnImageTap,
                         hasBackgroundImage)),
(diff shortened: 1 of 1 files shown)


w568w (approved):
LGTM. However, I still wonder why you removed the toolbars when the floors are deleted. The toolbar still displays useful information, even after being deleted!

HydrogenC:
> LGTM. However, I still wonder why you removed the toolbars when the floors are deleted. The toolbar still displays useful information, even after being deleted!

Due to inertia. The if statement as copied from the original code before the UI tweak.

PR #394, opened 2024-07-15
Audit Page

### LINKED ISSUE #372: [BUG] 审核界面循环出现相同内容
**描述 Bug**
审核界面请求api缺少order_by参数，导致当遇到某些帖子时，循环出现相同内容

**复现步骤**
复现该 Bug 的步骤：
1. 到 审核 页面
2. 不断下拉刷新，可能出现循环

**系统信息**
 - 设备型号：
 - 系统版本：
 - 旦夕版本（见「设置-关于本应用」）：1.4.3

kooWZ:
1. Closes #372
2. Show and highlight sensitive_detail in audit page
3. don't flush audit list after mark sensitive / not sensitive on a record. This is useful when checking old data

### DIFF
--- a/lib/repository/forum/forum_repository.dart
+++ b/lib/repository/forum/forum_repository.dart
@@ -678,7 +678,8 @@ class ForumRepository extends BaseRepositoryWithDio {
           "offset": startTime.toUtc().toIso8601String(),
           "size": length,
           "all": false,
-          "open": open
+          "open": open,
+          "order_by": "time_created"
         },
         options: Options(headers: _tokenHeader));
     final result = response.data;
--- a/lib/model/forum/audit.dart
+++ b/lib/model/forum/audit.dart
@@ -29,9 +29,17 @@ class OTAudit {
   final int modified;
   final String? time_created;
   final String? time_updated;
+  final String? sensitive_detail;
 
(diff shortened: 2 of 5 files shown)


lib/repository/forum/forum_repository.dart
```diff
@@ -678,7 +678,8 @@ class ForumRepository extends BaseRepositoryWithDio {
           "offset": startTime.toUtc().toIso8601String(),
           "size": length,
           "all": false,
-          "open": open
+          "open": open,
+          "order_by": "time_created"
```
w568w, line 682: Just out of curiosity (nothing to do with code review), what will happen if we do not set `order_by` here?

w568w (approved):
t8aENdiXyf

lib/repository/forum/forum_repository.dart
kooWZ: Not a big problem. Backend uses `time_created` by default if we don't set here. 
#372 is caused by https://github.com/kooWZ/DanXi/blob/20fc7af0dd9aa4b7a8e43ee235736519ce3ce2cd/lib/page/forum/hole_reports.dart#L283 sending `time_updated` as `time_created`

PR #403, opened 2024-08-08
feat: add version for forum answer question api

JingYiJun:


HydrogenC (changes requested):
I doubt if it's necessary to introduce a new dependency for returning multiple values. Using the builtin `record` language feature in Dart would be a better choice.  
Please refer to: https://dart.dev/language/records  
PS: This syntax is borrowed from C# actually

w568w:
> I doubt if it's necessary to introduce a new dependency for returning multiple values. Using the builtin `record` language feature in Dart would be a better choice.

Also, we have https://github.com/DanXi-Dev/DanXi/blob/main/lib/model/pair.dart for pre-`record` era.

HydrogenC:
> > I doubt if it's necessary to introduce a new dependency for returning multiple values. Using the builtin `record` language feature in Dart would be a better choice.
> 
> Also, we have https://github.com/DanXi-Dev/DanXi/blob/main/lib/model/pair.dart for pre-`record` era.

It isn't needed any more right now, is it?

HydrogenC (approved):

w568w:
> It isn't needed any more right now, is it?

Yeah, of course.

HydrogenC:
It looks good enough to me, so may we merge this right now @w568w ?

w568w (approved):
LGTM! Just let it go.

PR #412, opened 2024-08-31
add: scrolling for quiz option list

### LINKED ISSUE #408: [Feature Request] 注册答题界面增加一个下拉条
**你的功能需求和某个bug有关吗？**
`[否]`

**你想要什么样的功能？**
部分用户系统字体过大，无法选中全部选项，需要增加下拉条才可以完成答题

HydrogenC:
Fixes #408.  
Also fixed a long-existing bug in `PagedListView` that stuck me for over half an hour, demonstrated as followed:  

`_hasError` is set to `true` in `errorBuilder`, as we expect. 
<https://github.com/DanXi-Dev/DanXi/blob/c0037b9fb38b211e8649fc91de3d754855d9026b/lib/widget/libraries/paged_listview.dart#L223>
However, in the following lines, `_clearData` is called. 
<https://github.com/DanXi-Dev/DanXi/blob/c0037b9fb38b211e8649fc91de3d754855d9026b/lib/widget/libraries/paged_listview.dart#L228>
In which `_hasError` is restored back to `false`, overwriting the assignment on line 223. 
<https://github.com/DanXi-Dev/DanXi/blob/c0037b9fb38b211e8649fc91de3d754855d9026b/lib/widget/libraries/paged_listview.dart#L399-L404>
This causes `_hasError` to be `false` regardless of whether there is an error or not, which lead to unexpected behaviors when judging `_hasError` in `if` statements.

### DIFF
--- a/lib/widget/libraries/paged_listview.dart
+++ b/lib/widget/libraries/paged_listview.dart
@@ -225,7 +225,8 @@ class PagedListViewState<T> extends State<PagedListView<T>>
           if (snapshot.error != null &&
               snapshot.error is FatalException &&
               widget.fatalErrorBuilder != null) {
-            _clearData();
+            // We cannot clear the error here
+            _clearData(clearError: false);
             pageIndex = widget.startPage;
             return widget.fatalErrorBuilder!.call(context, snapshot.error);
           } else {
@@ -396,11 +397,13 @@ class PagedListViewState<T> extends State<PagedListView<T>>
   }
 
   /// Clear all the data saved.
-  void _clearData() {
+  void _clearData({bool clearError = true}) {
     _data.clear();
     valueKeys.clear();
-    _hasError = false;
     _dataClearQueued = false;
+    if(clearError){
+      _hasError = true;
+    }
   }
 
   @override
@@ -448,7 +451,7 @@ class PagedListViewState<T> extends State<PagedListView<T>>
     });
   }
 
-  replaceDatumWith(T data, int index){
+  replaceDatumWith(T data, int index) {
     setState(() {
       _data[index] = data;
     });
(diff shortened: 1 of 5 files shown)


lib/widget/libraries/paged_listview.dart
```diff
@@ -396,11 +397,13 @@ class PagedListViewState<T> extends State<PagedListView<T>>
...
   /// Clear all the data saved.
-  void _clearData() {
+  void _clearData({bool clearError = true}) {
     _data.clear();
     valueKeys.clear();
-    _hasError = false;
     _dataClearQueued = false;
+    if(clearError){
+      _hasError = true;
+    }
```
w568w, line 406: I don't get the logic here. If `_clearData()`'s `clearError` is default to be true, shouldn't any invocation of `_clearData()` except line 229 set `_hasError = true`? Is that what you want?

HydrogenC: I'm sorry, typo

lib/widget/libraries/paged_listview.dart
```diff
@@ -396,11 +397,13 @@ class PagedListViewState<T> extends State<PagedListView<T>>
   }
 
   /// Clear all the data saved.
-  void _clearData() {
+  void _clearData({bool clearError = true}) {
```
w568w, line 400: Just a suggestion: consider adding some comments here to explain why the argument `clearError` is necessary.

HydrogenC: Resolved

w568w (approved):
img_v3_02ef_a6259c4c-f731-4b25-b05d-08ea5e79c37g

PR #420, opened 2024-09-13
Fix #418

### LINKED ISSUE #418: [BUG] 研究生课表点击课程显示 Null 报错、左上角学期选择报错
**复现步骤**
复现该 Bug 的步骤：
1. 以研究生身份登录
2. 进入课表页面，刷新使课表出现
3. 观察到左上角「学期选择按钮」报错。点击任意课程
5. 出现 Null 错误

w568w:
This PR will:

1. fix #418;
2. fix some minor bugs and refactor a little.

PTAL at the commit history for details.

### DIFF
--- a/lib/page/home_page.dart
+++ b/lib/page/home_page.dart
@@ -17,6 +17,7 @@
 import 'dart:async';
 import 'dart:convert';
 
+import 'package:app_links/app_links.dart';
 import 'package:dan_xi/common/constant.dart';
 import 'package:dan_xi/common/pubspec.yaml.g.dart';
 import 'package:dan_xi/generated/l10n.dart';
@@ -27,9 +28,9 @@ import 'package:dan_xi/model/person.dart';
 import 'package:dan_xi/page/platform_subpage.dart';
 import 'package:dan_xi/page/subpage_danke.dart';
 import 'package:dan_xi/page/subpage_dashboard.dart';
+import 'package:dan_xi/page/subpage_forum.dart';
 import 'package:dan_xi/page/subpage_settings.dart';
 import 'package:dan_xi/page/subpage_timetable.dart';
-import 'package:dan_xi/page/subpage_forum.dart';
 import 'package:dan_xi/provider/forum_provider.dart';
 import 'package:dan_xi/provider/settings_provider.dart';
 import 'package:dan_xi/provider/state_provider.dart';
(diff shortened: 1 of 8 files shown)


HydrogenC (approved):
Question: shall we replace all `S.of(context)` with `S.current` in the current codebase later?

w568w:
> Question: shall we replace all `S.of(context)` with `S.current` in the current codebase later?

In practice: yes, you can. While using `S.current` might be considered a pattern that breaks design conventions, it generally won't cause significant issues.

The main difference is that `S.current` retrieves a "global" i18n instance, whereas `S.of(context)` gets the `Locale` of the parent widget in the widget tree. In most cases, an app only operates with a single locale at a time, so no major problems arise. This app is no exception.

That said, I still advocate for using `S.of(context)`. I appreciate the principle that **wherever you can access a `context`, the code in question is guaranteed to be refreshable. Otherwise, you're risking writing code that won't "rebuild" during the component's lifecycle.**

Consider this: when you can obtain a `context`, you're either inside a Widget's `build` method or in a builder function that provides a `cxt` argument. In both scenarios, you can be confident that the method will rerun whenever dependencies change.

Now, take a look at why I had to use `S.current` in this PR: I needed to fetch localized error messages in `initState`. This is problematic because `initState` only runs once during the lifecycle of the widget, regardless of any changes. In other words, **if the app's language changes, the error messages won't update accordingly**. You might argue that this isn't a big deal, since users typically won't switch languages while an error dialog is being displayed. But that's only true in this specific scenario. 

I prefer the way `S.of(context)` forces you to consider the context in which you're writing code, and what behavior that code will exhibit.

w568w:
By the way, I believe a more "elegant" solution to this problem would be to create a custom exception subclass that contains a message builder, for example:

```dart
// When throwing
throw LocalizedException(message: (cxt) => S.of(cxt).some_message);

// When catching and obtaining the message string
final message = e.getMessage(context);
```

HydrogenC:
> By the way, I believe a more "elegant" solution to this problem would be to create a custom exception subclass that contains a message builder, for example:
> 
> ```dart
> // When throwing
> throw LocalizedException(message: (cxt) => S.of(cxt).some_message);
> 
> // When catching and obtaining the message string
> final message = e.getMessage(context);
> ```

It's indeed quite a tricky problem for our current architecture. There's not a elegant and straightforward way to resolve this problem unless we adopt a context-independent library like GetX, and doing so might also resolve the warning of async functions accessing the context. Nevertheless I don't have that much time for massive bottom-to-up rewrite.

PR #429, opened 2024-10-27
Fixes #306

### LINKED ISSUE #42: [Feature Request] 导出考试日程
导出考试日程到ics

### LINKED ISSUE #306: [BUG] 课表添加周日课程时出现在选定周的前一周
**描述 Bug**
手动设置课表周日的课程时，假如只设置第10周，会出现在第9周的课表页面上

**复现步骤**
复现该 Bug 的步骤：
1. 到课表页面
2. 点击右上角手动添加课程
3. 勾选第10周开课，周日任意时间
4. 课程出现在第9周的课表页面

**系统信息**
 - 设备型号：跟设备没关系
 - 系统版本：跟系统没关系
 - 旦夕版本（见「设置-关于本应用」）：1.14.0 nightly build 336

AInfinity-LilacDream:
This PR adjust the courses logic, now courses pulled from jwc with lessons on Sunday will be saved a week forward to match Danxi's data structure, so that Sunday courses will be rendered correctly.

### DIFF
--- a/lib/repository/fdu/time_table_repository.dart
+++ b/lib/repository/fdu/time_table_repository.dart
@@ -92,12 +92,23 @@ class TimeTableRepository extends BaseRepositoryWithDio {
           "semester.id": await getAppropriateSemesterId()
         },
         options: DioUtils.NON_REDIRECT_OPTION_WITH_FORM_TYPE);
-    return TimeTable.fromHtml(
+    TimeTable timetable = TimeTable.fromHtml(
         startTime ??
             DateTime.tryParse(
                 SettingsProvider.getInstance().thisSemesterStartDate ?? "") ??
             Constant.DEFAULT_SEMESTER_START_DATE,
         tablePage.data!);
+    for (var course in timetable.courses!) {
+      for (var weekday in course.times!) {
+        if (weekday.weekDay == 6)  {
+          for (int i = 0; i < course.availableWeeks!.length; i++) {
+            course.availableWeeks![i] = course.availableWeeks![i] - 1;
+          }
+          break;
+        }
+      }
+    }
+    return timetable;
   }
 
   Future<TimeTable?> loadTimeTable(PersonInfo? info,
@@ -111,12 +122,25 @@ class TimeTableRepository extends BaseRepositoryWithDio {
           (cachedValue) => TimeTable.fromJson(jsonDecode(cachedValue!)),
           (object) => jsonEncode(object.toJson()));
     } else {
-      return Cache.get<TimeTable>(
+      Future<TimeTable?> timetable = Cache.get<TimeTable>(
           KEY_TIMETABLE_CACHE,
           () async =>
               (await loadTimeTableRemotely(info, startTime: startTime))!,
           (cachedValue) => TimeTable.fromJson(jsonDecode(cachedValue!)),
           (object) => jsonEncode(object.toJson()));
+      timetable.then((timetable) {
+        for (var course in timetable!.courses!) {
+          for (var weekday in course.times!) {
+            if (weekday.weekDay == 6)  {
+              for (int i = 0; i < course.availableWeeks!.length; i++) {
+                course.availableWeeks![i] = course.availableWeeks![i] - 1;
+              }
+              break;
+            }
+          }
+        }
+      });
+      return timetable;
     }
   }
 
--- a/lib/repository/fdu/time_table_repository.dart
+++ b/lib/repository/fdu/time_table_repository.dart
@@ -128,18 +128,6 @@ class TimeTableRepository extends BaseRepositoryWithDio {
               (await loadTimeTableRemotely(info, startTime: startTime))!,
           (cachedValue) => TimeTable.fromJson(jsonDecode(cachedValue!)),
           (object) => jsonEncode(object.toJson()));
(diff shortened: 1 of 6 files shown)


w568w (changes requested):
Let me further explain the current situation to help you understand the correct way to fix the issue. The timetable data flows through **three logical layers**:

1. **Timetable Retrieval Layer** (i.e., `repository/fdu/time_table_repository.dart`). This layer directly fetches the original webpage from the college server, parsing fields such as dates, and passing the results to the persistence layer.
2. **Timetable Persistence Layer** (i.e., `model/time_table.dart`). This layer is responsible for saving the data obtained from the retrieval layer in a standardized format, which is then used for storage on disk or for passing to the view layer for display.
3. **Timetable View Layer**. This layer includes all parts of the code where the timetable is used to implement the business logic (such as the timetable page, or the "next course" feature on the home page). These components retrieve the current timetable from `time_table.dart`.

The current problem is that, **for the Fudan undergraduate system, the schedule format received by the retrieval layer differs from our defined format, requiring an adjustment to the placement of Sunday courses in one of the layers above**. Previously, we performed this adjustment in **the persistence layer**.

However, for other academic systems, this adjustment might not be necessary and therefore should ideally be applied specifically in **the retrieval layer for the Fudan undergraduate system**.

Returning to your PR, it indeed changed this “adjustment”, but **in the opposite direction**. You moved the “adjustment” to the view layer, which doesn’t address the root of the issue.

w568w:
Let’s delve further into this issue:

In the Fudan system, each week starts on Sunday, whereas in DanXi, each week starts on Monday. For example, if September 1 is a Sunday and September 2 is a Monday:

1. For Fudan, the first week is:

```
September 1 (Sunday) through September 7 (Saturday).
```

2. For DanXi, however, the first week is:

```
September 2 (Monday) through September 8 (Sunday).
```
Now, suppose there is a math class scheduled for September 1. In Fudan’s system, it is recorded as `Week 1, Sunday` (Fudan version).

But if DanXi interprets it as `Week 1, Sunday` (DanXi version), **it will incorrectly display the course on September 8.**

To correct this, we want to adjust any Sunday courses immediately after retrieving the webpage from the Fudan system. This way, DanXi’s time_table will record this class as `Week 0, Sunday` (DanXi version), ensuring it displays correctly on September 1.

lib/repository/fdu/time_table_repository.dart
```diff
@@ -111,12 +122,25 @@ class TimeTableRepository extends BaseRepositoryWithDio {
...
           (object) => jsonEncode(object.toJson()));
     } else {
-      return Cache.get<TimeTable>(
+      Future<TimeTable?> timetable = Cache.get<TimeTable>(
           KEY_TIMETABLE_CACHE,
           () async =>
               (await loadTimeTableRemotely(info, startTime: startTime))!,
           (cachedValue) => TimeTable.fromJson(jsonDecode(cachedValue!)),
           (object) => jsonEncode(object.toJson()));
+      timetable.then((timetable) {
```
w568w, line 131: No need for `then()` here. To wait for a `Future`, simply use `await`, which also requires marking this method as `async`.

We originally didn't mark this method as `async` simply because it didn't require `await`. Now that it does, we can make that change.

w568w (changes requested):
Redacted. Please follow my latest review.

lib/repository/fdu/time_table_repository.dart
```diff
@@ -111,12 +122,25 @@ class TimeTableRepository extends BaseRepositoryWithDio {
...
+            if (weekday.weekDay == 6)  {
+              for (int i = 0; i < course.availableWeeks!.length; i++) {
+                course.availableWeeks![i] = course.availableWeeks![i] - 1;
+              }
+              break;
+            }
+          }
+        }
+      });
+      return timetable;
```
w568w, line 143: Two things to note here:

1. No need for `then()` here. To wait for a `Future`, simply use `await`, which also requires marking this method as `async`. We originally didn't mark this method as `async` simply because it didn't require `await`. Now that it does, we can make that change.
2. The handling is inappropriate in some cases. When calling `loadTimeTable()`, there are three different cases:
    1. **No timetable found locally, so fetch it using `loadTimeTableRemotely()`.** Here, the weekday issue has already been resolved, meaning **you’re adjusting it twice**.
    2. **A timetable is cached locally after the `loadTimeTableRemotely()` update.** In this case, the weekday issue has also been resolved, **so you’re still adjusting it twice**.
    3. **A timetable is cached locally from before the `loadTimeTableRemotely()` update (e.g., the user is upgrading from an older version of DanXi without re-fetching the timetable).** Here, the weekday issue has not been resolved, **so an adjustment is needed**.
    My suggestion is **NOT to adjust the timetable in this method**. Instead, include a note in the update log to remind users to manually refresh to get the corrected version. This approach eliminates case 3, so no adjustments are needed here.

w568w (changes requested):
Sorry for late review, just some minor changes.

Also, can you reword your commit message? It's still not following our guidelines. Check the [commit history](https://github.com/DanXi-Dev/DanXi/commits/main) for some examples.

w568w (approved):
LGTM! Thanks for your patience.

PR #447, opened 2024-11-25
Implement webvpn on Android

### LINKED ISSUE #362: [Feature Request] WebVPN 代理支持
**你想要什么样的功能？**
由于假期学生离校，需要接入 webvpn 代理以入校访问。

**如何实现？**
这不难实现，只需将请求的目标 URL 改为对应的 webvpn 版本即可。可以参考：

https://github.com/DanXi-Dev/DanXi-swift/blob/e5a2b542dc7a02f744f97b90d732740194161c36/Fudan%20Kit/Classroom/ClassroomAPI.swift#L57-L60

HydrogenC:
Closes #362. Tested on Windows.  
It works now. The only defect is that the waiting time for first load is very slow (because it waits too long for direct access). Even though I've set the timeout to 2s, it still takes ~10s for direct access attempt to fail, resulting in long loading time at first request.  
For reviewers: though I've checked already, please double-check that I have correctly set the request method of every api (I might have forgotten to change the request method after copying).  
Also fixes a bug where NullPointerEception will be raised when failed to acquire toml and the forum page turns red. Added some null checks so that this would no longer happen.


HydrogenC:
cc @kooWZ , please test if it works on Android.

lib/repository/base_repository.dart
```diff
@@ -40,13 +46,15 @@ abstract class BaseRepositoryWithDio {
       _dios[linkHost] = DioUtils.newDioWithProxy();
       _dios[linkHost]!.options = BaseOptions(
           receiveDataWhenStatusError: true,
-          connectTimeout: const Duration(seconds: 10),
-          receiveTimeout: const Duration(seconds: 10),
-          sendTimeout: const Duration(seconds: 10));
+          connectTimeout: const Duration(seconds: 2),
+          receiveTimeout: const Duration(seconds: 2),
+          sendTimeout: const Duration(seconds: 2));
```
w568w, line 51: Any special reason of tuning these timeout values?

lib/repository/danke/curriculum_board_repository.dart
```diff
@@ -66,74 +68,102 @@ class CurriculumBoardRepository extends BaseRepositoryWithDio {
...
+        await WebvpnProxy.requestWithProxy(dio, options);
     return response.statusCode;
   }
 
   Future<int?> modifyReview(
       int reviewId, CourseReviewEditorText updatedReview) async {
-    Response<String> response = await dio.put("$_BASE_URL/reviews/$reviewId",
+    final options = RequestOptions(
+        path: "$_BASE_URL/reviews/$reviewId",
+        method: "PUT",
```
w568w, line 124: Have you considered https://github.com/DanXi-Dev/DanXi/issues/362#issuecomment-2267446801?

lib/repository/forum/forum_repository.dart
```diff
@@ -497,47 +556,62 @@ class ForumRepository extends BaseRepositoryWithDio {
   }
 
   Future<OTUser?> updateUserProfile() async {
-    final Response<Map<String, dynamic>> response = await dio.put(
-        "$_BASE_URL/users/${provider.userInfo!.user_id}",
+    final options = RequestOptions(
+        path: "$_BASE_URL/users/${provider.userInfo!.user_id}",
+        method: "PUT",
```
w568w, line 561: Ditto.

lib/repository/base_repository.dart
```diff
@@ -40,13 +46,15 @@ abstract class BaseRepositoryWithDio {
...
       _dios[linkHost]!.interceptors.add(LimitedQueuedInterceptor.getInstance());
       _dios[linkHost]!.interceptors.add(UserAgentInterceptor(
           userAgent: SettingsProvider.getInstance().customUserAgent));
-      _dios[linkHost]!.interceptors.add(CookieManager(cookieJar!));
+      // _dios[linkHost]!.interceptors.add(CookieManager(cookieJar!));
+      // Cookies related with webvpn
+      _dios[linkHost]!.interceptors.add(CookieManager(globalCookieJar));
```
w568w, line 57: I’m quite concerned about this change: have you considered scenarios where WebVPN is not enabled? 

If this change is correct, it would mean that the CookieJar isolation we previously implemented was completely unnecessary (this opinion clearly requires more thorough testing), and you should also remove `_cookieJars` completely.

Otherwise, if it’s incorrect, have you considered cases where cookies need to be separated? For example, `fdty.fudan.edu.cn` uses a login method different from normal UIS login, so we previously used a separate `cookieJar` to handle their login processes individually.

Additionally, I’m also concerned about codes where the "local" cookieJar is cleared arbitrarily, such as with `cookieJar.clearAllCookies()`… Anyway, I think this change will require more time for thorough testing.

lib/util/webvpn_proxy.dart
```diff
@@ -0,0 +1,158 @@
...
+
+    return true;
+  }
+
+  /// Guard after
+  static Future<void> loginWebVpn(Dio dio) async {
+    if (!isLoggedIn) {
+      // Another concurrent task is running
+      if (loginSession != null) {
+        await loginSession;
```
w568w, line 89: Just a comment: A cleverer solution to prevent concurrent login! 👍 

I believe I have made it too complex in `UISLoginTool`, since I am a little old-school. 😅 Maybe we should refactor that some day.

lib/util/webvpn_proxy.dart
```diff
@@ -0,0 +1,158 @@
...
+        return;
+      }
+
+      debugPrint("Logging into WebVPN");
+      loginSession = UISLoginTool.loginUIS(
+          dio,
+          WebvpnProxy.WEBVPN_LOGIN_URL,
+          BaseRepositoryWithDio.globalCookieJar,
+          StateProvider.personInfo.value);
```
w568w, line 98: Technically, we do not allow the use of the global `StateProvider` instance outside the application layer... But well, perhaps `WebvpnProxy` could be considered part of the application layer in some sense. A more compliant solution would be to add a `PersonInfo` parameter to this function, but this might make higher-level calls to `WebvpnProxy` too complex.

lib/util/webvpn_proxy.dart
```diff
@@ -0,0 +1,158 @@
...
+    await loginWebVpn(dio);
+
+    // Second attempt
+    response = await dio.fetch<T>(options);
+    if (checkResponse(response)) {
+      return response;
+    }
+
+    // All attempts failed
+    throw WebVpnLoginException(options.method);
```
w568w, line 156: Can we make the exception message more self-descriptive? An `"Error: POST"` could be confusing for debugging.

w568w (changes requested):
Great work! I've just completed some preliminary static code analysis and might conduct more in-depth testing later.

w568w:
By the way, I think we might need to squash and clean up the commit history. Right now, it contains some experimental commits and merge commits. I can help with that after you've completed all the requested changes from the review.

lib/repository/danke/curriculum_board_repository.dart
HydrogenC: Handled in the inplementation of `requestWithProxy`.

kooWZ:
> cc @kooWZ , please test if it works on Android.

I will check this as soon as server maintainance is over

lib/util/webvpn_proxy.dart
HydrogenC: I decoupled this by adding a `_personInfo` field in the class, and the field will be initialized when loading UIS info into memory.

lib/repository/base_repository.dart
HydrogenC: To shorten the timeout for direct connect, or we'll have to wait for too long during first connect.

HydrogenC:
@w568w I created a new `ReadonlyCookieJar` and a separate side-by-side cookie manager for resolving the problem. Also, I added a field `isWebvpnApplicable` for each linkhost so that the webvpn cookie jar will only be added to dios related with forum and cirricular board. But I'm not sure if I've done this correctly, please review.

lib/repository/base_repository.dart
HydrogenC: See comment below.

lib/page/home_page.dart
```diff
@@ -632,6 +630,9 @@ class HomePageState extends State<HomePage> with WidgetsBindingObserver {
       LoginDialog.showLoginDialog(
           context, preferences, StateProvider.personInfo, false);
     }
+
+    // Forward the person info to webvpn services
+    WebvpnProxy.initPerson(StateProvider.personInfo.value);
```
w568w, line 635: `LoginDialog.showLoginDialog` does not block! So when the user hasn't logged in, you are accessing `StateProvider.personInfo.value` before it's set.

lib/util/webvpn_proxy.dart
```diff
@@ -0,0 +1,176 @@
...
+      options.validateStatus = (status) {
+        return status != null && status < 400;
+      };
+    }
+
+    // See: https://github.com/DanXi-Dev/DanXi/issues/362#issuecomment-2267446801
+    if (options.method == "PUT"){
+      options.method = "PATCH";
+      options.path += "/_webvpn";
+    }
```
w568w, line 152: Consider adding notes for this path hacking behavior in the docstring of `requestWithProxy`?

A careless developer might use the PUT method in a repository for other websites and be confused by unexpected behavior (maybe 404 Not Found). Alternatively, maybe you might simply remove these lines and manually change all PUT requests to PATCH in Forum and Danke repository, or add a method argument to control the hacking behavior.

lib/repository/base_repository.dart
```diff
@@ -34,19 +37,25 @@ abstract class BaseRepositoryWithDio {
...
+          connectTimeout: const Duration(seconds: 2),
+          receiveTimeout: const Duration(seconds: 2),
+          sendTimeout: const Duration(seconds: 2));
       _dios[linkHost]!.interceptors.add(LimitedQueuedInterceptor.getInstance());
       _dios[linkHost]!.interceptors.add(UserAgentInterceptor(
           userAgent: SettingsProvider.getInstance().customUserAgent));
       _dios[linkHost]!.interceptors.add(CookieManager(cookieJar!));
+      if (isWebvpnApplicable && SettingsProvider.getInstance().useWebvpn) {
+        _dios[linkHost]!.interceptors.add(CookieManager(WebvpnProxy.webvpnCookieJar));
+      }
```
w568w, line 58: According to [the source](https://github.com/cfug/dio/blob/540616204d73814fd07855ef0e0a3b88424ec0a5/plugins/cookie_manager/lib/src/cookie_mgr.dart#L60-L61), if you add multiple `CookieManager`s to a dio instance, The last added `CookieManager` will always overwrite the entire `Cookie` field in the header. So it is equivalent to only adding `CookieManager(WebvpnProxy.webvpnCookieJar)`.

If you want both to make effects, you may need kind of `CookieManagerMerger`…

w568w (changes requested):

lib/util/webvpn_proxy.dart
HydrogenC: But I still have to keep PUT as-is when I'm using LAN instead of WebVPN, haven't I?

w568w: > But I still have to keep PUT as-is when I'm using LAN instead of WebVPN, haven't I?

No! Our server certainly does not stop one from using the `PATCH` variant in the LAN environment.

lib/repository/base_repository.dart
HydrogenC: Well, that's pretty discouraging, but I could handle this.

lib/page/home_page.dart
HydrogenC: Tricky point indeed.

HydrogenC: Resolved in [e420abf](https://github.com/DanXi-Dev/DanXi/pull/447/commits/e420abf4604ebddaf1b417aa812bc51447188f50)

lib/util/webvpn_proxy.dart
w568w: > So it would be more preferrable to both use the PATCH api regardless of the connect mode?

Yes.

HydrogenC:
Resolved.

w568w (commented):
Mostly LGTM! We need to go through some real-world tests before merging.

w568w:
Two problems I noticed when testing on Windows:

1. The initial request is still slow, taking ~5 seconds or more.
2. Forum images fail to load. It might be worth considering proxying image requests as well. See <https://github.com/DanXi-Dev/DanXi/blob/main/lib/util/io/cache_manager_with_proxy.dart> for inspiration.

w568w (changes requested):
Two problems I noticed when testing on Windows:

1. The initial request is still slow, taking ~5 seconds or more. It needs to be fixed in this PR.
2. Forum images fail to load. It might be worth considering proxying image requests as well. See <https://github.com/DanXi-Dev/DanXi/blob/main/lib/util/io/cache_manager_with_proxy.dart> for inspiration. Since this requires additional work, my suggestion is to submit an issue immediately after this PR is merged to track the progress, ensuring it doesn't delay the release.

HydrogenC:
> 1. The initial request is still slow, taking ~5 seconds or more. It needs to be fixed in this PR.
The only method is to decrese the timeout when trying direct links. 
> 2. Forum images fail to load. It might be worth considering proxying image requests as well. See https://github.com/DanXi-Dev/DanXi/blob/main/lib/util/io/cache_manager_with_proxy.dart for inspiration. Since this requires additional work, my suggestion is to submit an issue immediately after this PR is merged to track the progress, ensuring it doesn't delay the release.
Why is this happening, I have already proxied IMAGE_BASE_URL. Is there any other modules that have to be changed to make it work?

w568w:
> The only method is to decrese the timeout when trying direct links.

Or, you can heuristically determine whether one is on campus:

https://github.com/DanXi-Dev/DanXi/blob/56c29c8570332ea352460d5691fdce9dc5bb7872/lib/page/subpage_dashboard.dart#L103-L111

> Is there any other modules that have to be changed to make it work?

The problem is that the `CachedNetworkImage` widget does not rely on Dio at all. It requests images through `DefaultCacheManager` that declares a `fileService`, which contains `HttpClient`. You may have to re-route that `HttpClient`. Check the file I listed above.

HydrogenC:
> > The only method is to decrese the timeout when trying direct links.
> 
> Or, you can heuristically determine whether one is on campus:

No, this approach is also slow. The off-campus notification pops up after ～5s after application launch. If the url isn't available, dio won't return before counting all the timeout. 

Sidenote: I wouldn't be able to code until tonight due to experiment courses. Sorry for high latency of resolving code reviews.

HydrogenC:
As for the first point, it now responds within 1~2 seconds.  

As for the second point, fixed and tested on windows (image below, taken under webvpn mode)
屏幕截图 2024-11-27 163240

w568w:
dc5671a3f3b3e8e85e70b69e45d5c94bd4346c0a removes supports of #369, do I get it right?

HydrogenC:
> Edit: I am wrong. `HttpFileServiceWithWebvpn` uses `DioUtils.newDioWithProxy` to create a dio supporting HTTP proxy. Never mind.

Yeah I mentioned this in the comment of `DefaultCacheManagerWithWebvpn`.

lib/util/webvpn_proxy.dart
```diff
@@ -0,0 +1,249 @@
...
+  String toString() {
+    if (message == null) return "Login to WebVPN has failed";
+    return "Login to WebVPN has failed: $message";
+  }
+}
+
+class WebvpnProxy {
+  // These async session objects work like a semaphore, which only allow one concurrent action and blocks all other requesters until the action completes
+  static Future<void>? loginSession;
+  static Future<bool>? tryDirectSession;
```
w568w, line 25: (This is just a comment)

I would say that the most similar thing coming to my mind is a mutex with `try_acquire`, because I will accomplish the same behavior in C/C++ with something like:

```cpp
if (lock.try_acquire()) {
    // ... do logging in ...
} else {
    lock.acquire(); // only block and wait for finishing
}
lock.release();
```

w568w (approved):
Impressive. Will merge it as soon as I complete some testing.

lib/util/webvpn_proxy.dart
HydrogenC: Okay, I considered using mutex before, however it doesn't support `try_acquire` so I gave up.

w568w: > Okay, I considered using mutex before, however it doesn't support `try_acquire` so I gave up.

That's why I utilized a bunch of things in `UISLoginTool`: read-write lock, epoch-based task, etc. 😅

w568w:
I have tested on Android. Everything works.

PR #316, opened 2023-10-31
Fix #226 and upgrade dependencies

w568w:
This PR fixes #226 and upgrade dependencies.

I also applied `dart fix` to adapt to the recently introduced `super.xxx` grammar.

**Since I have changed some key dependencies and NOT walked through a thorough test, please check**:

- Whether saving to gallery still works on mobile platforms (i.e. Android, iOS)
- Whether the storage location (or keys?) of shared preferences have changed (there is a way to check this: upgrade from previous build, finding that all settings and accounts are lost, then the shared preferences should be read from different location and the answer is YES)

PR #449, opened 2024-11-26
Upgrade secure storage and fix regression issue introduced by migration logic

w568w:
This PR is a follow-up to #316. It:

**1. upgrades `flutter_secure_storage` and `encrypt_shared_preferences` to their latest versions.**

In testing on Windows, `flutter_secure_storage` did not encounter any issues during the upgrade. **However, since it has not been tested on Android and Linux, compatibility remains unclear.**

`encrypt_shared_preferences` was previously reported to break backward compatibility, because the author silently changed the default encryption algorithm parameters. This PR corrects the parameters to their original values, ensuring that data from previous versions can be decrypted correctly.

**2. fixes a regression**

Fix the issue where incorrect detection of data migration status caused data to become invalid after logging out, logging back in, and then restarting the application. The reproducing steps are:

1. Go to Settings, log out of the UIS account, and select "Yes" when prompted to "Delete all data."  
2. Log back into the UIS account and go to the homepage.  
3. Exit the app and clear it from the background.  
4. Reopen the app.  
5. At this point, you should be prompted to log in to UIS again, as if the login credential from step 2 was not remembered.


HydrogenC (approved):
Looks nice and works fine on Windows.

w568w:
Since this PR fixes some issue, move it into the milestone.

ivanfei-1 (approved):
Tested, LGTM.

PR #473, opened 2025-01-01
Fix "only this person"

w568w:
This PR:

1. fixes the problem that when filtering posts by a specified user only in the deeper levels of the thread, the post page refreshes itself and becomes blank;
2. fixes the problem that the status description and actual logic of the "show this user" are inconsistent: when a user filters by the poster of a certain post, long-pressing on another post displays "Show all posts," but clicking it actually filters by the user of that other post.

---

However, there's still an operation that can easily trap users: if a user filters by the poster of a deeper comment and then pulls down to refresh, the entire page becomes blank because the pull-to-refresh only fetches the first 10 posts. Perhaps the filter setting should be reset upon pull-to-refresh?

Perhaps the best solution would be to automatically fetch the next 10 posts on pull-to-refresh until posts satisfying the filter are found. However, this sounds like it would put a lot of pressure on the backend.

### DIFF
--- a/lib/page/forum/hole_detail.dart
+++ b/lib/page/forum/hole_detail.dart
@@ -334,12 +334,10 @@ class BBSPostDetailState extends State<BBSPostDetail> {
                       setState(() {
                         if ((_renderModel as Normal).selectedPerson != hole.floors?.first_floor?.anonyname) {
                           (_renderModel as Normal).selectedPerson = hole.floors?.first_floor?.anonyname;
-                        }
-                        else {
+                        } else {
                           (_renderModel as Normal).selectedPerson = null;
                         }
                       });
-                      refreshListView();
                     }),
                 PopupMenuOption(
                     label: _multiSelectMode
@@ -887,16 +885,15 @@ class BBSPostDetailState extends State<BBSPostDetail> {
       PlatformContextMenuItem(
         menuContext: menuContext,
         onPressed: () async {
-          if ((_renderModel as Normal).selectedPerson != null) {
-            (_renderModel as Normal).selectedPerson = null;
-            refreshListView();
-          } else {
-            setState(() =>
-            (_renderModel as Normal).selectedPerson = e.anonyname);
-            refreshListView();
-          }
+          setState(() {
+            if ((_renderModel as Normal).selectedPerson != null) {
+              (_renderModel as Normal).selectedPerson = null;
+            } else {
+              (_renderModel as Normal).selectedPerson = e.anonyname;
+            }
+          });
         },
-        child: Text((_renderModel as Normal).selectedPerson == e.anonyname ? S.of(context).show_all_replies : S.of(context).show_this_person),
+        child: Text((_renderModel as Normal).selectedPerson != null ? S.of(context).show_all_replies : S.of(context).show_this_person),
       ),
       PlatformContextMenuItem(
         menuContext: menuContext,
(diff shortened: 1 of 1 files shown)


AInfinity-LilacDream:
Is it possible that when only this person triggers, we do not pull-to-refresh, but instead we record posts that have been fetched and set their, say, attribute to invisible to users?

w568w:
> Is it possible that when only this person triggers, we do not pull-to-refresh, but instead we record posts that have been fetched and set their, say, attribute to invisible to users?

This was done in https://github.com/DanXi-Dev/DanXi/pull/473/commits/3cb7ca1785a56bea4ab1ec72f9965a1a3741b5c5. I removed the `refreshListView()` calls so it wouldn't trigger a refresh when clicking "filtering by this person".

AInfinity-LilacDream:
"Perhaps the best solution would be to automatically fetch the next 10 posts on pull-to-refresh until posts satisfying the filter are found. However, this sounds like it would put a lot of pressure on the backend."
I think this may be a method but isn't practical. a solution may be quit the "only this person" mode when refreshing.

AInfinity-LilacDream (commented):
if you consider my suggestion as impractical, then this pr should be good.

AInfinity-LilacDream:
sorry for the late comment.

AInfinity-LilacDream (approved):

AInfinity-LilacDream:
this should work.

lib/page/forum/hole_detail.dart
```diff
@@ -68,7 +68,7 @@ String preprocessContentForDisplay(String content) {
     if (element is UrlElement) {
       // Only add tag if tag has not yet been added.
       if (RegExp("\\[.*?\\]\\(${RegExp.escape(element.url)}\\)")
-              .hasMatch(content) ||
+          .hasMatch(content) ||
```
w568w, line 71: **Please avoid including irrelevant code formatting in your commits.** 

Each commit should ideally contain only the necessary changes. As we don't currently have a code formatting guideline, if your IDE automatically formats code (e.g., randomly adding/removing spaces, tabs, newlines), please review your file changes carefully before committing. 

Both VSCode and Android Studio allow you to commit specific lines within a file. This enables you to include only the intended changes in your current commit, deferring other formatting adjustments to subsequent commits (or reverting them).

lib/page/forum/hole_detail.dart
```diff
@@ -203,7 +203,10 @@ class BBSPostDetailState extends State<BBSPostDetail> {
   }
 
   /// Refresh the list view.
-  Future<void> refreshListView({bool scrollToEnd = false}) async {
+  Future<void> refreshListView({bool scrollToEnd = false, required selectedPerson, required VoidCallback func}) async {
+    if (!scrollToEnd && selectedPerson != null) {
+      func();
+    }
```
w568w, line 209: This strikes me as rather confusing and over-abstracted. 

If you're passing in a function callback and a `selectedPerson` parameter, simply to check a condition and then conditionally execute that callback, **wouldn't it make more sense to perform that check directly at the point where this function is invoked?** Especially considering there's currently **only one instance in the entire codebase where an actually useful value is being passed in, and the logic within the callback bears absolutely no relation to refreshing.**

What's your rationale for putting this code – which, to me, looks completely non-reusable and seems to have no bearing on what `refreshListView` is supposed to do – inside the `refreshListView` method?

w568w (commented):
Given the changes in 8f88a74879631b606de66ddf5445fc63dc5cb379, which I found quite puzzling, I'm going to submit a request to revert that commit.

PR #474, opened 2025-01-03
Revert "fix: now when user pulls to refresh while in "only this person" mode, it will automatically quit." and fix it

w568w:
This reverts commit 8f88a74879631b606de66ddf5445fc63dc5cb379.

Also, this PR isn't solely for reverting the commit; it also aims to fix it and properly address the issue mentioned in #473. I'm marking it as a draft for now since this part isn't complete yet.

### DIFF
--- a/lib/page/forum/hole_detail.dart
+++ b/lib/page/forum/hole_detail.dart
@@ -418,6 +418,7 @@ class BBSPostDetailState extends State<BBSPostDetail> {
                 backgroundColor: Theme.of(context).dialogBackgroundColor,
                 onRefresh: () async {
                   HapticFeedback.mediumImpact();
+                  (_renderModel as Normal).selectedPerson = null;
                   await refreshListView();
                 },
                 child: pagedListView),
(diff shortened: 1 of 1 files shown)


w568w (commented):
This commit really shouldn't be called a "refactor" – it doesn't actually involve any refactoring, which is a software-engineering thing. 

If you've accidentally committed the wrong changes and messed up the commit history, a better way to handle it is to fix the Git history locally and then force push to completely remove the commit(s) that caused the mess. 

While this isn't an option in many repositories, it's acceptable to do in this project. (In repositories where force pushing is prohibited, the best practice is to close the PR, clean up the history, put the changes on a new branch, and open a fresh PR with the new branch.)

lib/page/forum/hole_detail.dart
```diff
@@ -418,6 +418,7 @@ class BBSPostDetailState extends State<BBSPostDetail> {
                 backgroundColor: Theme.of(context).dialogBackgroundColor,
                 onRefresh: () async {
                   HapticFeedback.mediumImpact();
+                  (_renderModel as Normal).selectedPerson = null;
```
w568w, line 421: 1. you should add some comments explaining the reason for inserting this line of code;
2. I'm thinking that maybe when in "view only original poster" mode, a refresh shouldn't reset the state, since the original poster will always be on the first page!

w568w (commented):
@AInfinity-LilacDream

AInfinity-LilacDream:
this should work. @w568w

w568w:
LGTM. Thanks for the efforts!

PR #476, opened 2025-01-04
Fix #475

### LINKED ISSUE #475: [BUG] 水印只应出现在树洞和蛋壳
如题。

w568w:
This PR:

1. implements reference counting for `Watermark`, allowing it to automatically hide when no longer required by any pages or widgets;
2. only sends `ViewStateChanged.INVISIBLE` event to the tab page that was just switched away from, to ensure that each page receives a matching and equal number of `VISIBLE` and `INVISIBLE` events.

Fixes #475.


Boreas618:
LGTM

PR #483, opened 2025-01-10
Reformat all codes and apply dart fix

w568w:
We have never definitively addressed the code style issues in our project. I believe now is the time to ***thoroughly*** resolve them.

The problems caused by inconsistent code styles are self-evident. For example, every time contributors submit a PR to this project, they may unintentionally modify unrelated lines of code (adding or removing spaces or reformatting line breaks) because their formatting tools "casually" reformatted the entire file.

I think it's necessary to establish a unified code style. This isn't particularly challenging since Dart's default formatter doesn't even support customization. In theory, all Dart projects should have exactly the same code style.

That said, I still urge everyone to check their formatters:

1. **Avoid using unusual third-party formatters**. The Dart plugins for VSCode and Android Studio should only invoke the `dartfmt` tool for formatting.
2. **Ensure the code width is set to 80 characters**. This is the default width recommended by Flutter's official guidelines. In theory, your IDE should already default to this width when running `dartfmt`, but please double-check it. In Android Studio, you can find this setting under `File | Settings | Editor | Code Style | Dart`.

In addition, please make sure to **run code formatting every time before committing**. Moving forward, I may also consider setting up GitHub Actions to check whether the changes in each PR comply with the formatting requirements.

### DIFF
--- a/build_release.dart
+++ b/build_release.dart
@@ -114,7 +114,8 @@ Future<int> runDartProcess(List<String> args) async {
   return await buildProcess.exitCode;
 }
 
-Future<void> buildAndroid(String? versionCode, String gitHash, {String? target}) async {
+Future<void> buildAndroid(String? versionCode, String gitHash,
+    {String? target}) async {
   print('Build for Android...');
   await runFlutterProcess([
     'build',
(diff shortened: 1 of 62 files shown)


w568w:
As this does no harm, I am going to merge it directly.

PR #484, opened 2025-01-12
Remove `bitsdojo_window` dependency

w568w:
The `bitsdojo_window` package was previously used to control window maximization and minimization. However, we've removed this functionality since 17e83792d25eadd31bc84638c43f042a8ee28206.

Currently, it serves no purpose and hasn't been maintained for >2 years.

This pull request will remove it. If similar functionality is needed in the future, the actively maintained [`window_manager`](https://pub.dev/packages/window_manager) package can be used instead.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -18,7 +18,6 @@
 import 'dart:async';
 import 'dart:ui';
 
-import 'package:bitsdojo_window/bitsdojo_window.dart';
 import 'package:dan_xi/feature/feature_map.dart';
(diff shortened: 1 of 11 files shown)


ivanfei-1:
I’m extremely happy that we finally got rid of this package, as it has been a pain in the ass in terms of dependency and version management.

w568w:
> I’m extremely happy that we finally got rid of this package

Also, I've just realized that the `tray_manager` dependency is no longer needed either. I'll submit a pull request to remove it later.

PR #486, opened 2025-01-13
feat: use `OTFloorWidget` in modify history

kooWZ:
This allows for showing & jumping to mentions in modify history, which can help administrators make better judgments based on the context.

### DIFF
--- a/lib/page/forum/admin_operation.dart
+++ b/lib/page/forum/admin_operation.dart
@@ -5,7 +5,6 @@ import 'package:dan_xi/provider/settings_provider.dart';
 import 'package:dan_xi/repository/forum/forum_repository.dart';
 import 'package:dan_xi/util/master_detail_view.dart';
 import 'package:dan_xi/util/noticing.dart';
-import 'package:dan_xi/util/forum/human_duration.dart';
 import 'package:dan_xi/util/platform_universal.dart';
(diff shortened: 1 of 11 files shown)


lib/page/forum/admin_operation.dart
```diff
@@ -161,11 +160,12 @@ class AdminOperationPageState extends State<AdminOperationPage> {
...
-                        itemBuilder: (e) => Card(
-                            child: ListTile(
-                                subtitle: Text(HumanDuration.tryFormat(context,
-                                    DateTime.tryParse(e.time_updated ?? ""))),
-                                title: Text(e.content!))),
+                        itemBuilder: (e) => OTFloorWidget(
+                            hasBackgroundImage: false,
+                            floor: _floor.copyWith(
+                                content: e.content,
+                                time_updated: e.time_updated,
```
w568w, line 167: We don't actually use or display `time_updated` in the `OTFloorWidget`. It might be better to switch to `time_created`.

w568w (changes requested):

lib/page/forum/admin_operation.dart
kooWZ: Fixed

w568w (approved):
LGTM.

PR #488, opened 2025-01-23
fix: webvpn login

ivanfei-1:


lib/util/io/dio_utils.dart
```diff
@@ -62,6 +66,9 @@ class DioUtils {
       if (!Uri.parse(location).isAbsolute) {
         location = '${response.requestOptions.uri.origin}/$location';
       }
+      if (jar != null) {
+        dio.interceptors.add(CookieManager(jar));
```
w568w, line 70: This really doesn't make sense, I am afraid.

First, `dio.interceptors.add` isn't idempotent. Calling `processRedirect` multiple times will add multiple `CookieManager`s, and that's definitely not expected.

Secondly, there's no obvious connection between `processRedirect` and cookie management. Why are you passing an `IndependentCookieJar` to it, and why is `processRedirect` responsible for adding a `CookieManager`? This action is completely unrelated to what `processRedirect` should be doing and it breaks the abstraction. It’s like that, you see function A and B happen to be executed together in the code, and then you directly copy-pastes B's content into A's function body – there's no clear reason behind it. I'm really confused.

lib/util/io/dio_utils.dart
```diff
@@ -21,6 +21,9 @@ import 'package:dan_xi/provider/settings_provider.dart';
 import 'package:dan_xi/util/platform_universal.dart';
 import 'package:dio/dio.dart';
 import 'package:dio/io.dart';
+import 'package:dio_cookie_manager/dio_cookie_manager.dart';
+
+import '../../repository/cookie/independent_cookie_jar.dart';
```
w568w, line 26: We don't allow using relative imports in our codes. Please change them to package imports (`import 'package:..../foo.dart'`).

lib/util/webvpn_proxy.dart
```diff
@@ -1,12 +1,22 @@
...
+import 'package:dan_xi/util/platform_universal.dart';
+import 'package:dan_xi/util/public_extension_methods.dart';
 import 'package:dio/dio.dart';
+import 'package:dio5_log/interceptor/diox_log_interceptor.dart';
+import 'package:dio_cookie_manager/dio_cookie_manager.dart';
+import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
 
+import 'io/queued_interceptor.dart';
+import 'io/user_agent_interceptor.dart';
```
w568w, line 18: Ditto.

lib/util/webvpn_proxy.dart
```diff
@@ -119,21 +126,121 @@ class WebvpnProxy {
...
+    if (response.statusCode != 200) {
+      throw Exception(
+          "Failed to log in to WebVPN, status code: ${response.statusCode}");
+    }
+    webvpnCookieJar.cloneFrom(webvpnJar);
+    isLoggedIn = true;
+  }
+
+  // w568w's [UISLoginTool.loginUIS] and [UISLoginTool.tryAsyncWithAuth] are not reusable and comprehensible in any case ;(
+  // so I rewrite this function here
```
w568w, line 195: Who is "I"? Please leave your name here for the future review, e.g.:

```dart
// @ivanfei (2025-01-20): I rewrite this function here
```

lib/util/webvpn_proxy.dart
```diff
@@ -119,21 +126,121 @@ class WebvpnProxy {
...
+
+    if (response.statusCode != 200) {
+      throw Exception(
+          "Failed to log in to WebVPN, status code: ${response.statusCode}");
+    }
+    webvpnCookieJar.cloneFrom(webvpnJar);
+    isLoggedIn = true;
+  }
+
+  // w568w's [UISLoginTool.loginUIS] and [UISLoginTool.tryAsyncWithAuth] are not reusable and comprehensible in any case ;(
```
w568w, line 194: Could you clarify the difference between the `_authenticateWithUIS` method and `UISLoginTool.loginUIS`? 

I've compared them and haven't been able to spot any significant differences.

lib/util/webvpn_proxy.dart
```diff
@@ -119,21 +126,121 @@ class WebvpnProxy {
...
+      return null;
+    }
+
+    BeautifulSoup soup = BeautifulSoup(response.data!);
+
+    for (final element in soup.findAll('input')) {
+      if (element.attributes['name'] == 'ticket') {
+        return element.attributes['value'];
+      }
+    }
```
w568w, line 238: It can be simplified to `soup.find('input', attrs: {'name': 'ticket'})`. Check the [documentation](https://github.com/mzdm/beautiful_soup?tab=readme-ov-file#usage).

lib/util/webvpn_proxy.dart
```diff
@@ -119,21 +126,121 @@ class WebvpnProxy {
...
+            await DioUtils.processRedirect(idDio, firstResponse, idJar);
+
+        final location = await _authenticateWithUIS();
+        final Response<dynamic> responseWithTicket = await idDio.get(location);
+
+        String? ticket = _retrieveTicket(responseWithTicket);
+        if (ticket == null) {
+          throw Exception("Failed to retrieve ticket from id.fudan.edu.cn");
+        }
+        await _performWebvpnLogin(ticket);
```
w568w, line 162: Why not use `UISLoginTool.tryAsyncWithAuth` here? As far as I can see, you are:

1. Retrieve a ticket from `https://id.fudan.edu.cn/idp/authCenter/authenticate?service=https%3A%2F%2Fwebvpn.fudan.edu.cn%2Flogin%3Fcas_login%3Dtrue`. If success, finish with `_performWebvpnLogin`; otherwises go to step 2;
2. Login UIS at `https://uis.fudan.edu.cn/authserver/login?service=https://id.fudan.edu.cn/idp/thirdAuth/cas`. If success, retry step 1 again; otherwises throw an exception.

These two steps are ***exactly*** what `UISLoginTool.tryAsyncWithAuth` was designed for. Why are you reinventing the wheels here (and you don't handle every possible exception correctly)?

w568w (commented):
There are really something confusing me. Please explain them or I would rather submit a PR to rollback these changes.

ivanfei-1:
I would glad to explain throughly, but I also have to point out that rolling back does not solve issues, PRs do.

w568w:
> I would glad to explain throughly, but I also have to point out that rolling back does not solve issues, PRs do.

I would like to clarify that, reverting the changes != rejection of them. Feel free to explore your ideas or address the issues in a ***separate*** branch. However, I don't think this PR is quite up to par for merging into master at this stage.

PR #504, opened 2025-03-09
Fix webvpn relative URL

w568w:
This PR addresses the issue where the `isResponseRequiringLogin` function was incorrectly determining the necessity for re-login, due to overlooking the possibility that `response.realUri` could be a relative URL.

However, this fix does not fully resolve the problem where, after being forcibly logged out of WebVPN due to an IP address change and subsequently logging back in, the initial request invariably redirects to `/login`.  From the second request onward, the system responds as expected, even without requiring another login. We still need to investigate this problem further.


w568w:
I'm going to go ahead and merge this PR as it introduces a minor change and is essentially harmless. (Besides, I might be the only one on this project who understands this part anyway...)

PR #505, opened 2025-03-09
Update deps

w568w:
This PR updates `flutter_platform_widgets` to version 8.0.0, as it includes support for Flutter 3.29, which is the version we are currently using.


ivanfei-1 (approved):
LGTM.

PR #510, opened 2025-03-18
feat #440: add pe test score to pe_feature page

AInfinity-LilacDream:
This is a temporary solution, since I think that maybe an individual page should look better. If you consider this version as good, then that's ok. If you consider my thought as reasonable, I will write a new page in a few days.

### DIFF
--- a/lib/feature/pe_feature.dart
+++ b/lib/feature/pe_feature.dart
@@ -29,7 +29,7 @@ import 'package:flutter/material.dart';
 
 class PEFeature extends Feature {
   PersonInfo? _info;
-  List<dynamic>? _exercises;
+  List<ExerciseObjects>? _exercises;
 
   /// Status of the request.
   ConnectionStatus _status = ConnectionStatus.NONE;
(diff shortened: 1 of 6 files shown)


lib/repository/fdu/pe_repository.dart
```diff
@@ -72,16 +95,33 @@ class FudanPERepository extends BaseRepositoryWithDio {
 class ExerciseItem {
   final String title;
   final int? times;
+  final String type;
```
w568w, line 98: ... and remove `type`s here and there.

It's worth noting that while this is valid in TypeScript (where TypeScript suggests using Literal Types to differentiate objects, e.g., `function foo(alignment: "left" | "right" | "center")`), different languages have their own ways of handling this. 

In Swift, Haskell, Dart, and Rust, you should make good use of ADTs.

lib/feature/pe_feature.dart
```diff
@@ -29,7 +29,7 @@ import 'package:flutter/material.dart';
 
 class PEFeature extends Feature {
   PersonInfo? _info;
-  List<ExerciseItem>? _exercises;
+  List<dynamic>? _exercises;
```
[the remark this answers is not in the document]
AInfinity-LilacDream: has replaced with ADT.

lib/repository/fdu/pe_repository.dart
AInfinity-LilacDream: removed.

lib/feature/pe_feature.dart
```diff
@@ -75,43 +75,45 @@ class PEFeature extends Feature {
...
-              case '加章1':
-                exerciseCategory[0] += element.times!;
-                break;
-              case '加章2':
-                exerciseCategory[1] += element.times!;
-                break;
-              case '加章3':
-                exerciseCategory[2] += element.times!;
-                break;
+            if (element.type == "Item") {
```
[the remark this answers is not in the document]
AInfinity-LilacDream: replaced.

lib/repository/fdu/pe_repository.dart
```diff
@@ -92,36 +92,35 @@ class FudanPERepository extends BaseRepositoryWithDio {
...
       list.add(ExerciseItem(elements[i].text.trim().replaceFirst("：", ""),
-          int.tryParse(elements[i + 1].text.trim()), "Item"));
+          int.tryParse(elements[i + 1].text.trim())));
     }
     return list;
   }
 }
 
-class ExerciseRecord {
+class ExerciseRecord implements ExerciseObjects {
```
w568w, line 114: I suppose the singular form (`Object`) sounds better than plural ones (`Objects`)?

AInfinity-LilacDream: replaced.

w568w (approved):
This looks good, but I have a few suggestions regarding the commit messages:

*   `upgrade: add riverpod`: "upgrade" is typically used for dependency upgrades.  Here, `refactor` would be more appropriate.
*   `fix: replace dynamic with ADTS`: "fix" refers to bug fixes in specific application functionalities.  This commit obviously rewrites the code structure (without changing the logic) and does NOT fix any bugs, so `refactor` should be used.
*   `chore: replace plural name 'ExerciseObjects' with singular form`: "chore" refers to changes in the build process and is unrelated to variable names. You should use `style` instead.

PR #516, opened 2025-04-23
feat: Implement Dynamic Color, Refactor Theming & Navigation

Bengerthelorf:
This commit introduces dynamic color support based on system palettes and addresses related UI/logic issues.

- Add `dynamic_color` package dependency (pubspec.yaml, pubspec.lock).
- Register `dynamic_color` plugin for desktop platforms (generated_plugin_registrant).
- Add new Android launcher icons (foreground and monochrome).

- Refactor theme building logic in `main.dart`:
    - Integrate `DynamicColorBuilder` to fetch system color schemes.
    - Use `context.watch` to listen for `SettingsProvider` changes.
    - Correctly apply dynamic ColorScheme when 'followSystemPalette' is enabled, fixing incomplete theme application.

- Add 'followSystemPalette' setting:
    - Update `SettingsProvider` to manage the new setting.
    - Add relevant localization strings (en, ja, zh).

- Refactor Settings page (`subpage_settings.dart`):
    - Disable manual theme color selection when 'followSystemPalette' is true.
    - Remove `FlutterApp.restartApp` calls after theme/palette changes to fix crashes during rebuild.

- Fix missing AppBar/NavigationBar on pushed subpages:
    - Modify `PlatformSubpageState` (`platform_subpage.dart`) to remove the `_isInTab` check, ensuring all subpages consistently build their own scaffold and app bar.

- Update Navigation:
    - Add navigation to Settings page from Dashboard AppBar (`subpage_dashboard.dart`).
    - (Verify if Settings was also removed from bottom nav in `home_page.dart` as auto-commit suggested).

- Other:
    - Improve robustness of `SettingsProvider.primarySwatch` getter.


android/app/src/main/res/drawable/launch_background.xml
w568w, line 1: Why removing these icons from launch splash screen?

android/gradle.properties
w568w, line 1: Typically, you don't need to do this because Flutter explicitly specifies the JDK location when running Gradle, which overrides any other settings (`org.gradle.java.home`, `JAVA_HOME`, `PATH`).

See https://github.com/flutter/flutter/pull/128264 and [this article](https://medium.com/@mahmoudnabil14/demystifying-flutter-configurations-jdk-gradle-and-fvm-with-flutter-sidekick-f0dde5ed058a) for details.

You can use `flutter doctor -v` to check the JDK that is used at the moment.

lib/main.dart
```diff
@@ -225,68 +225,108 @@ class DanxiApp extends StatelessWidget {
...
-                  "ERROR: onUnknownRoute() has been called inside the root navigator.\nDevelopers are not supposed to push on this Navigator. There should be something wrong in the code."),
-              home: ThemedSystemOverlay(
-                child: PlatformMasterDetailApp(
-                  // Configure the page route behaviour of the whole app.
-                  onGenerateRoute: (settings) {
-                    final Function? pageContentBuilder =
+        return DynamicColorBuilder(
+          builder: (ColorScheme? lightDynamic, ColorScheme? darkDynamic) {
+            // Use context.watch to listen for setting changes
+            final settings = context.watch<SettingsProvider>();
```
w568w, line 231: The problem with `watch` is that it triggers a rebuild whenever `SettingsProvider` notifies of ***any*** update. 

Is there a particular reason you're using it instead of `context.select`, which only triggers a rebuild when the ***specific*** property you're interested in changes?

lib/main.dart
```diff
@@ -225,68 +225,108 @@ class DanxiApp extends StatelessWidget {
...
+            bool isPlatformCupertino = PlatformX.isCupertino(context);
+
+            // Get base themes for copying non-color properties if needed
+            // Pass a default color like Colors.blue, it won't be used if dynamic is active
+            ThemeData baseLightTheme = Constant.lightTheme(isPlatformCupertino, manualPrimarySwatch);
+            ThemeData baseDarkTheme = Constant.darkTheme(isPlatformCupertino, manualPrimarySwatch);
+
+            if (useSystemPalette && lightDynamic != null && darkDynamic != null) {
+              // Build themes directly using the dynamic ColorScheme
+              lightThemeConfig = ThemeData(
```
w568w, line 250: Use `baseLightTheme.copyWith` instead, if you want to copy the base theme with several fields overrided.

lib/page/home_page.dart
```diff
@@ -706,12 +704,12 @@ class HomePageState extends State<HomePage> with WidgetsBindingObserver {
...
-                      ? const Icon(Icons.settings)
-                      : const Icon(CupertinoIcons.gear_alt),
-                  label: S.of(context).settings,
-                ),
+                // BottomNavigationBarItem(
+                //   icon: PlatformX.isMaterial(context)
+                //       ? const Icon(Icons.settings)
+                //       : const Icon(CupertinoIcons.gear_alt),
+                //   label: S.of(context).settings,
+                // ),
```
w568w, line 712: This appears to be a design choice. What was the rationale for moving the Settings tab entry to the top right corner of the homepage?

lib/page/subpage_dashboard.dart
```diff
@@ -62,12 +63,20 @@ class HomeSubpage extends PlatformSubpage<HomeSubpage> {
...
+        Icon(Icons.create_outlined),
             () => smartNavigatorPush(cxt, '/dashboard/reorder').then(
-                (value) => RefreshHomepageEvent(onlyRefreshOrder: true).fire()))
-      ];
+                (value) => RefreshHomepageEvent(onlyRefreshOrder: true).fire())),
+
+    AppBarButtonItem(
+        S.of(cxt).settings,
+        Icon(Icons.settings_outlined),
+            () => Navigator.of(cxt).push(
```
w568w, line 75: Use `smartNavigatorPush` unless you have a good reason to avoid it.

lib/provider/settings_provider.dart
```diff
@@ -657,16 +658,24 @@ class SettingsProvider with ChangeNotifier {
...
+
     if (preferences!.containsKey(KEY_PRIMARY_SWATCH_V2)) {
       int? color = preferences!.getInt(KEY_PRIMARY_SWATCH_V2);
-      return Color(color!).value;
+      if (color != null) {
+        return Color(color).value;
+      } else {
+      }
+    } else {
     }
```
w568w, line 672: Remove these empty `else` branches.

lib/provider/settings_provider.dart
```diff
@@ -657,16 +658,24 @@ class SettingsProvider with ChangeNotifier {
 
   /// Primary color used by the app.
   int get primarySwatch {
+    if (preferences == null) {
+      return Colors.blue.value;
+    }
```
w568w, line 663: `preferences` should never be null since it's initialized in `main()`, before `runApp()`. 

You can safely remove this guard.

lib/provider/settings_provider.dart
```diff
@@ -657,16 +658,24 @@ class SettingsProvider with ChangeNotifier {
...
+
     if (preferences!.containsKey(KEY_PRIMARY_SWATCH_V2)) {
       int? color = preferences!.getInt(KEY_PRIMARY_SWATCH_V2);
-      return Color(color!).value;
+      if (color != null) {
+        return Color(color).value;
```
w568w, line 668: Just `return value` is enough, as what you change below.

lib/page/subpage_settings.dart
```diff
@@ -36,7 +35,7 @@ import 'package:dan_xi/util/io/cache_manager_with_webvpn.dart';
 import 'package:dan_xi/util/master_detail_view.dart';
 import 'package:dan_xi/util/noticing.dart';
 import 'package:dan_xi/util/platform_universal.dart';
-import 'package:dan_xi/util/public_extension_methods.dart';
+// import 'package:dan_xi/util/public_extension_methods.dart';
```
w568w, line 38: You can just remove the unused imports.

w568w (changes requested):
Good first PR! There are some changes that you may pay attention to.

w568w:
Also, I'd like to note that `subpage_settings` is a horrible monster, cluttered with various bad practices from early days. 

It's not that your refactoring is wrong, but compared to refactoring, a rewrite is actually more practical. We plan to rewrite the entire file using `Riverpod` later. If you're interested, feel free to come talk to me about it in detail.

android/app/src/main/res/drawable/launch_background.xml
Bengerthelorf: Removing the splash icon because it looks blurry on my test Samsung S9 and I want to hide it. On Android 12+, the Splash Screen API generates the icon anyway, so it has no negative impact, and this also provides a cleaner launch screen.

lib/main.dart
```diff
@@ -150,6 +151,7 @@ class DanxiApp extends StatelessWidget {
     '/placeholder': (context, {arguments}) => ColoredBox(
         color: Theme.of(context).scaffoldBackgroundColor.withAlpha(254)),
     '/home': (context, {arguments}) => const HomePage(),
+    '/settings': (context, {arguments}) => const SettingsPage(),
```
w568w, line 155: Still, you haven't told me why you feel necessary to move setting into a separate page?

lib/main.dart
```diff
@@ -225,68 +227,101 @@ class DanxiApp extends StatelessWidget {
...
+                child: PlatformApp(
+                  // Remember? We have just defined this scroll behavior class above
+                  // to enable scrolling with mouse & stylus.
+                  scrollBehavior: TouchMouseScrollBehavior(),
+                  debugShowCheckedModeBanner: false,
+                  // Fix cupertino UI text color issue by override text color
+                  cupertino: (context, __) => CupertinoAppData(
+                      theme: CupertinoThemeData(
+                          brightness: context
+                              .select<SettingsProvider, ThemeType>( //
```
w568w: Remove the `//` here.

lib/main.dart
```diff
@@ -225,68 +227,101 @@ class DanxiApp extends StatelessWidget {
...
+                          textTheme: CupertinoTextThemeData(
+                              textStyle: TextStyle(
+                                  color: (Theme.of(context).brightness == Brightness.dark // Use current theme brightness
+                                      ? Colors.white
+                                      : Colors.black)
+                              )))),
+                  material: (context, __) => MaterialAppData(
+                    // Pass the ThemeData determined by DynamicThemeController
+                    theme: Theme.of(context),
+                  ),
```
w568w, line 288: `PlatformX.getTheme` is still being used in `widget/feature_item/feature_card_item.dart`. 

If you feel that `PlatformX.getTheme` is unnecessary, please also remove usage in that file and remove the method completely.

lib/page/subpage_dashboard.dart
```diff
@@ -62,12 +62,18 @@ class HomeSubpage extends PlatformSubpage<HomeSubpage> {
 
   @override
   Create<List<AppBarButtonItem>> get trailing => (cxt) => [
-        AppBarButtonItem(
-            S.of(cxt).dashboard_layout,
-            Text(S.of(cxt).edit, textScaler: TextScaler.linear(1.2)),
+    AppBarButtonItem(
+        S.of(cxt).dashboard_layout,
+        Icon(Icons.create_outlined),
```
w568w, line 67: Still, why change it to an icon? Is the text too large for you?

lib/page/subpage_settings.dart
w568w: You still didn't remove it!

lib/page/subpage_settings.dart
```diff
@@ -314,6 +311,7 @@ class SettingsSubpageState extends PlatformSubpageState<SettingsSubpage> {
     List<Widget> list = [];
     onTapListener(ThemeType theme) {
       SettingsProvider.getInstance().themeType = theme;
+      // setState(() {});
```
w568w: Remove the commented codes if they are unused.

lib/provider/settings_provider.dart
```diff
@@ -657,16 +658,12 @@ class SettingsProvider with ChangeNotifier {
 
   /// Primary color used by the app.
   int get primarySwatch {
-    if (preferences!.containsKey(KEY_PRIMARY_SWATCH_V2)) {
-      int? color = preferences!.getInt(KEY_PRIMARY_SWATCH_V2);
-      return Color(color!).value;
-    }
-    return Colors.blue.value;
+    return preferences!.getInt(KEY_PRIMARY_SWATCH_V2) ?? Colors.blue.value;
```
w568w: `Color(color!).value` is actually a validator to check whether the int `value` is a valid color hex value. I would rather not remove it.

lib/page/subpage_settings.dart
```diff
@@ -72,17 +70,14 @@ Future<void> updateOTUserProfile(BuildContext context) async {
...
+class SettingsPage extends StatefulWidget {
+  const SettingsPage({super.key});
 
   @override
-  Create<Widget> get title => (cxt) => Text(S.of(cxt).settings);
+  State<SettingsPage> createState() => _SettingsPageState();
 }
 
-class SettingsSubpageState extends PlatformSubpageState<SettingsSubpage> {
+class _SettingsPageState extends State<SettingsPage> {
```
w568w: Should be `SettingsPageState` (without `_`). 

According to the code style, the state class shouldn't be a private class. Linter would warn you about it.

lib/page/subpage_settings.dart
```diff
@@ -327,288 +322,322 @@ class SettingsSubpageState extends PlatformSubpageState<SettingsSubpage> {
...
-                            })),
+                            }
+                        ),
+
+                            if (PlatformX.isAndroid)
+                              SwitchListTile.adaptive(
+                                title: Text(S.of(context).follow_system_palette),
+                                subtitle: Text(S.of(context).follow_system_palette_description),
+                                secondary: const Icon(Icons.palette_outlined),
+                                value: SettingsProvider.getInstance().followSystemPalette,
```
w568w: DO NOT switch to use `SettingsProvider.getInstance()` in the `build()`. It is an anti-pattern.

Uniformly utilize `context.read` and `context.select` for reading and listening to properties.

w568w (changes requested):
Some last and final changes.

lib/common/constant.dart
```diff
@@ -387,7 +387,6 @@ class Constant {
       cardTheme: CardThemeData(
         elevation: 0.5,
         margin: const EdgeInsets.fromLTRB(10, 8, 10, 8),
-        color: ThemeData.light().cardColor,
```
w568w, line 390: Why remove the `color: ThemeData.light().cardColor`?

lib/page/subpage_settings.dart
w568w: You are still using `SettingsProvider.getInstance().followSystemPalette`!

lib/page/subpage_settings.dart
```diff
@@ -327,288 +322,323 @@ class SettingsSubpageState extends PlatformSubpageState<SettingsSubpage> {
   }
 
   @override
-  Widget buildPage(BuildContext context) {
+  Widget build(BuildContext context) {
+    final bool supportsDynamicColor = context.watch<SettingsProvider>().supportsDynamicColor;
```
w568w: 1. You'd better NOT use `context.watch<SettingsProvider>()`. 
  As I said, it listens to the WHOLE `SettingsProvider` and when ANY field in `SettingsProvider` changes, it will refresh the page. It is mostly not what you want.
  To listen to only the target field, use `context.select`.
2. `supportsDynamicColor` doesn't magically become listenable. You should call `notifyListener()` when it changes to notify the listeners. Since you didn't, the `context.select` will never be triggered correctly. If this is what you intended, you'd better use `context.read` to obtain a value without setting up listening.

Bengerthelorf:
> Why remove the `color: ThemeData.light().cardColor`?

It hardcodes the color. The code takes a static, fixed color from Flutter's default light theme. It does not use the dynamic ColorScheme. And That's the reason why the cards here are all pink

Bengerthelorf:
Thanks for the great suggestion. You're right, since `supportsDynamicColor` is determined at startup and won't change, there's no need to listen for updates. I've changed it to `context.read` which is more appropriate here. Thanks again for pointing this out!

w568w (approved):
LGTM. Thanks for your work!

PR #541, opened 2025-06-15
chore: use aspectRatio instead of width in tablet detection

Integral-Tech:
IMHO in tablet detection `aspectRatio` is better than `width`.


w568w (approved):
LGTM.

PR #560, opened 2025-07-28
feat: implement cloud sticker feature (#446)

KYLN24:
- Add CloudSticker model with id, url, and sha256 fields
- Integrate sticker management into AnnouncementRepository
- Support loading stickers from TOML configuration
- Implement SHA256 validation and local caching
- Update sticker picker UI to show both local and cloud stickers
- Add automatic sync on app startup

The stickers are loaded directly from the main TOML file using the [[sticker]] array format. Each sticker entry contains:
- id: sticker identifier
- url: download URL
- sha256: hash for integrity verification

🤖 Generated with [Claude Code](https://claude.ai/code)


lib/model/cloud_sticker.dart
```diff
@@ -0,0 +1,46 @@
...
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import 'package:json_annotation/json_annotation.dart';
+
+part 'cloud_sticker.g.dart';
+
+@JsonSerializable()
+class CloudSticker {
```
w568w: Better called `RemoteSticker` or `NetworkSticker`.

lib/page/forum/hole_editor.dart
```diff
@@ -343,53 +342,84 @@ class BBSEditorWidgetState extends State<BBSEditorWidget> {
...
                     ),
                   ]),
             ),
           );
-          return PlatformX.isCupertino(context)
-              ? ConstrainedBox(
-                  constraints: BoxConstraints(
-                      maxHeight: 0.66 * MediaQuery.of(context).size.height),
-                  child: Card(child: body))
-              : body;
```
w568w: Why remove the `PlatformX.isCupertino` logic here?

lib/provider/cloud_sticker_provider.dart
w568w, line 1: This provider is not used anywhere. Why create it?

lib/repository/app/announcement_repository.dart
```diff
@@ -15,19 +15,26 @@
...
 import 'package:dan_xi/util/io/dio_utils.dart';
 import 'package:dan_xi/util/shared_preferences.dart';
 import 'package:dio/dio.dart';
+import 'package:path_provider/path_provider.dart';
 import 'package:pub_semver/pub_semver.dart';
 import 'package:toml/toml.dart';
 
 class AnnouncementRepository {
   static const KEY_SEEN_ANNOUNCEMENT = "seen_announcement";
+  static const String _CACHE_KEY_STICKERS = "cloud_sticker_cache";
```
w568w: No, it's meaningless to store `key`s in the shared preferences. Just scan the cache directory during startup and generate sticker caches on the fly.

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +195,172 @@ class AnnouncementRepository {
 
     return _tomlCache!['highlight_tag_ids'].cast<int>();
   }
+
+  // Cloud Sticker Methods
+  Future<String> get _stickerCacheDir async {
+    if (_cacheDirectory != null) return _cacheDirectory!;
+    final dir = await getApplicationSupportDirectory();
```
w568w: Should be `getApplicationCacheDirectory()`.

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +195,172 @@ class AnnouncementRepository {
...
+  Future<List<CloudSticker>> loadStickersFromNetwork() async {
+    // Ensure TOML is loaded
+    if (_tomlCache == null) {
+      await loadAnnouncements();
+    }
+    
+    // Get stickers directly from the main TOML file
+    final stickers = getStickersFromCache();
+    return stickers ?? [];
+  }
```
w568w: Again, it's meaningless.

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +195,172 @@ class AnnouncementRepository {
...
+    if (_tomlCache == null) {
+      await loadAnnouncements();
+    }
+    
+    final cached = await getCachedStickers();
+    if (cached.isEmpty) {
+      return await syncStickers();
+    }
+    return cached;
+  }
```
w568w: **What the hell** did it create _3 different versions_ of `getStickers`? Shouldn't it be just ONE `getStickers` exposed? 

My brain refuses to review the garbage here, which wrapped the same logic like Russian nesting dolls

lib/util/stickers.dart
```diff
@@ -51,3 +54,18 @@ String? getStickerAssetPath(String stickerName) {
...
+
+Future<String?> getStickerPath(String stickerName) async {
+  final localAssetPath = getStickerAssetPath(stickerName);
+  if (localAssetPath != null) {
+    return localAssetPath;
+  }
+  
+  final repository = AnnouncementRepository.getInstance();
+  final cloudPath = await repository.getStickerFilePath(stickerName);
+  if (cloudPath != null && File(cloudPath).existsSync()) {
```
w568w: ??? Use `existsSync` in an `async` method. WTF

lib/widget/forum/render/render_impl.dart
```diff
@@ -105,14 +106,34 @@ final kMarkdownRenderFactory = (double? defaultFontSize) =>
...
-              width: 50,
-              height: 50,
-            );
-          }
+          return FutureBuilder<String?>(
+            future: getStickerPath(url),
+            builder: (context, snapshot) {
+              if (snapshot.hasData && snapshot.data != null) {
+                final path = snapshot.data!;
+                if (path.startsWith('assets/')) {
```
w568w: Shouldn't use the fragile logic (`.startsWith('assets/')`) to determine between `assets` and `file`.

The best practice is to make `getStickerPath` return a sum ADT.

lib/main.dart
```diff
@@ -110,6 +112,12 @@ void main() {
       final registerDeviceIdentity =
           PlatformX.isAndroid ? DeviceIdentity.register() : Future.value();
       registerDeviceIdentity.then((_) {
+        // Initialize cloud stickers on app startup (background)
+        // Only sync if sticker URL is available
+        AnnouncementRepository.getInstance().loadAnnouncements().then((_) {
+          unawaited(AnnouncementRepository.getInstance().syncStickers());
+        });
+
```
w568w: Fuck it. Did the AI really read the code and know where this should be put? Did it know that `loadAnnouncements` had been called elsewhere and this change would update the data ***twice*** each startup?

lib/main.dart
```diff
@@ -357,7 +365,8 @@ class DanxiApp extends StatelessWidget {
       child: MultiProvider(providers: [
         ChangeNotifierProvider.value(value: SettingsProvider.getInstance()),
         ChangeNotifierProvider(create: (_) => NotificationProvider()),
-        ChangeNotifierProvider.value(value: fduHoleProvider)
+        ChangeNotifierProvider.value(value: fduHoleProvider),
+        ChangeNotifierProvider(create: (_) => CloudStickerProvider())
```
w568w: Ditto. It's meaningless.

w568w (changes requested):

w568w:
Garbage.

lib/page/forum/hole_editor.dart
```diff
@@ -374,42 +373,74 @@ class BBSEditorWidgetState extends State<BBSEditorWidget> {
...
-                                    fit: BoxFit.contain,
-                                  ),
-                                ),
-                              );
-                            }).toList(),
-                          ),
-                        ),
+                      child: FutureBuilder<List<RemoteSticker>>(
+                        future: AnnouncementRepository.getInstance().getAvailableStickers(),
+                        builder: (context, remoteSnapshot) {
```
w568w: Should distinguish async states (Success / Error / Loading) here, and return different widgets accordingly.

lib/page/forum/hole_editor.dart
```diff
@@ -374,42 +373,74 @@ class BBSEditorWidgetState extends State<BBSEditorWidget> {
...
+                                            height: 60,
+                                            child: const Center(
+                                              child: CircularProgressIndicator(strokeWidth: 2),
+                                            ),
+                                          );
+                                        },
+                                      ),
+                                    ),
+                                  );
+                                }).toList(),
```
w568w: Ditto. Return an error image widget (i.e. a red exclamation mark on grey background. Check the image widget for examples) if any error.

lib/util/stickers.dart
```diff
@@ -15,39 +15,29 @@
...
-  dx_confused,
-  dx_like;
+import 'dart:io';
+import 'package:dan_xi/repository/app/announcement_repository.dart';
+
+sealed class StickerPath {
+  const StickerPath();
+}
+
+class AssetStickerPath extends StickerPath {
```
w568w: Add some comments to tell the reader why this class is not being used (i.e. for future use)

lib/widget/forum/render/render_impl.dart
```diff
@@ -106,14 +107,33 @@ final kMarkdownRenderFactory = (double? defaultFontSize) =>
...
+                  ),
+                };
+              }
+              return const SizedBox(
+                width: 50,
+                height: 50,
+                child: Center(
+                  child: CircularProgressIndicator(),
+                ),
+              );
```
w568w: Ditto. Or if you believe `getStickerPath` never throws, explicitly document that in the code comment of `getStickerPath`.

lib/main.dart
```diff
@@ -110,6 +111,9 @@ void main() {
       final registerDeviceIdentity =
           PlatformX.isAndroid ? DeviceIdentity.register() : Future.value();
       registerDeviceIdentity.then((_) {
+        // Initialize remote stickers on app startup (background)
+        unawaited(AnnouncementRepository.getInstance().syncStickers());
```
w568w, line 115: Better move it to `HomePage`.

lib/repository/app/announcement_repository.dart
```diff
@@ -37,6 +43,7 @@ class AnnouncementRepository {
 
   factory AnnouncementRepository.getInstance() => _instance;
   Map<String, dynamic>? _tomlCache;
+  String? _cacheDirectory;
```
w568w: Rename to a more descriptive name, e.g. `_remoteStickerCacheDir`.

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +193,141 @@ class AnnouncementRepository {
...
+    return _cacheDirectory!;
+  }
+
+
+
+  Future<List<String>> getCachedStickerIds() async {
+    final cacheDir = await _stickerCacheDir;
+    final directory = Directory(cacheDir);
+
+    if (!directory.existsSync()) {
```
w568w: Do NOT use `xxSync` in an async method!

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +193,141 @@ class AnnouncementRepository {
...
+      return [];
+    }
+
+    return directory
+        .listSync()
```
w568w: Ditto.

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +193,141 @@ class AnnouncementRepository {
...
+      final fileName = file.path.split(Platform.pathSeparator).last;
+      return fileName.substring(
+          0, fileName.length - 5); // Remove .webp extension
+    }).toList();
+  }
+
+  Future<String?> getStickerFilePath(String stickerId) async {
+    final cacheDir = await _stickerCacheDir;
+    final file = File("$cacheDir/$stickerId.webp");
+    return file.existsSync() ? file.path : null;
```
w568w: Ditto.

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +193,141 @@ class AnnouncementRepository {
...
+  Future<String> _calculateSha256(Uint8List data) async {
+    final digest = sha256.convert(data);
+    return digest.toString();
+  }
+
+  Future<bool> _validateStickerFile(
+      String filePath, String expectedSha256) async {
+    try {
+      final file = File(filePath);
+      if (!file.existsSync()) return false;
```
w568w: Ditto. What the hell did it write?

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +193,141 @@ class AnnouncementRepository {
...
+      final actualSha256 = await _calculateSha256(data);
+      return actualSha256 == expectedSha256;
+    } catch (e) {
+      return false;
+    }
+  }
+
+  Future<bool> downloadAndValidateSticker(RemoteSticker sticker) async {
+    try {
+      final response = await DioUtils.newDioWithProxy().get<Uint8List>(
```
w568w: Do NOT create a new HTTP client each time downloading a sticker. Cache it in the class.

lib/repository/app/announcement_repository.dart
```diff
@@ -186,6 +193,141 @@ class AnnouncementRepository {
...
+
+      final data = await file.readAsBytes();
+      final actualSha256 = await _calculateSha256(data);
+      return actualSha256 == expectedSha256;
+    } catch (e) {
+      return false;
+    }
+  }
+
+  Future<bool> downloadAndValidateSticker(RemoteSticker sticker) async {
```
w568w: If this method is only used in annoucement_repository, make it private.

lib/util/stickers.dart
KYLN24: Claude's response:

Status: ✅ Dead code has been cleaned up

  After investigation, I found that the AssetStickerPath class was indeed dead        
  code. While it was referenced in switch expressions in two files
  (render_impl.dart and hole_editor.dart), it was never actually instantiated -       
  the getStickerPath() function only returns FileStickerPath or null.

  Changes made:
  - Removed the unused AssetStickerPath class from lib/util/stickers.dart
  - Updated switch expressions to directly cast to FileStickerPath
  - Simplified the sealed class hierarchy to remove the unused variant

  Result: The code is now cleaner and more maintainable, with no unused classes or    
   unreachable code paths. All functionality remains intact since only
  FileStickerPath was ever returned by the actual logic.

  The dead code has been successfully removed, eliminating the confusion between      
  what was defined vs. what was actually used.

lib/provider/state_provider.dart
```diff
@@ -36,6 +38,34 @@ class StateProvider {
 
   static String? onlineUserAgent;
 
+  /// Available stickers loaded from the server.
+  static final ValueNotifier<List<RemoteSticker>?> availableStickers = ValueNotifier(null);
+  static final ValueNotifier<bool> stickersLoading = ValueNotifier(false);
+  static final ValueNotifier<dynamic> stickersError = ValueNotifier(null);
```
w568w, line 45: **No, no, no, no, no.** This is the **worst possible implementation** I can think of, for two reasons:

1. The value, loading, and error properties are states that have a **sum** relationship, but here they are expressed as a **product**. **This is a completely incorrect state abstraction.** It serves no purpose other than to introduce more meaningless states into the state space (for example, a state where `loading = true` and error is not null).
2. `StateProvider` is essentially a runtime global variable store for the application. Global variables are generally an anti-pattern and should be avoided whenever possible. I see absolutely no reason to add more global variables, especially when the alternative is to place them in more semantically relevant locations like `AnnouncementRepository` or `RemoteSticker`. Instead, they are forced into a completely unrelated class. This will be a disaster for future maintenance and readability.

lib/provider/state_provider.dart
```diff
@@ -36,6 +38,34 @@ class StateProvider {
...
+    
+    try {
+      final stickers = await AnnouncementRepository.getInstance().getAvailableStickers();
+      availableStickers.value = stickers;
+    } catch (error) {
+      stickersError.value = error;
+    } finally {
+      stickersLoading.value = false;
+    }
+  }
```
w568w: And this next piece of code sends things from "very bad" to "**hellish**." 

I was speechless when I saw business logic placed directly inside the `StateProvider`. I don't think the AI understood a single byte of the project's architecture.

w568w:
TL;DR: it's even worse than garbage.

PR #572, opened 2025-09-01
Undergraduate timetable

kooWZ:
this commit also:
- Ensure `timetableLastUpdated` and `thisSemesterStartDate` are set on first load
- Handle case where default semester ID is not found in timetable (e.g., default is summer semester but student did not take summer courses)

Related to #542.

### DIFF
--- a/lib/page/subpage_timetable.dart
+++ b/lib/page/subpage_timetable.dart
@@ -169,7 +169,7 @@ class TimetableSubPageState extends PlatformSubpageState<TimetableSubPage> {
           UserGroup.FUDAN_UNDERGRADUATE_STUDENT) {
         _contentFuture = LazyFuture.pack(Retrier.runAsyncWithRetry(() =>
             TimeTableRepository.getInstance().loadTimeTable(
-                StateProvider.personInfo.value,
+                StateProvider.personInfo.value!,
                 forceLoadFromRemote: forceLoadFromRemote)));
       } else if (forceLoadFromRemote) {
         _contentFuture = LazyFuture.pack(
@@ -614,11 +614,14 @@ class SemesterSelectionButtonState extends State<SemesterSelectionButton> {
     String? chosenSemester = SettingsProvider.getInstance().timetableSemester;
     if (chosenSemester == null || chosenSemester.isEmpty) {
       chosenSemester = _semesterInfo!.defaultSemesterId;
+      SettingsProvider.getInstance().timetableSemester = chosenSemester;
     }
     _selectionInfo = _semesterInfo!.semesters
         .firstWhere((element) => element.semesterId == chosenSemester!);
     SettingsProvider.getInstance().semesterStartDates =
         _semesterInfo!.startDates;
+    SettingsProvider.getInstance().thisSemesterStartDate =
+        _semesterInfo!.startDates.parseStartDate(chosenSemester!);
   }
 
   @override
@@ -650,8 +653,10 @@ class SemesterSelectionButtonState extends State<SemesterSelectionButton> {
                               SettingsProvider.getInstance()
                                   .semesterStartDates
                                   ?.parseStartDate(
-                                      StateProvider.personInfo.value!.group,
-                                      e.semesterId!);
+                                    e.semesterId!,
+                                    group:
+                                        StateProvider.personInfo.value!.group,
+                                  );
                           if (parsedStartDate != null) {
                             SettingsProvider.getInstance()
                                 .thisSemesterStartDate = parsedStartDate;
--- a/lib/page/subpage_timetable.dart
+++ b/lib/page/subpage_timetable.dart
@@ -597,7 +597,7 @@ class SemesterSelectionButton extends StatefulWidget {
 }
 
 class SemesterSelectionButtonState extends State<SemesterSelectionButton> {
-  TimeTableSemesterInfo? _semesterInfo;
+  SemesterBundle? _semesterBundle;
   SemesterInfo? _selectionInfo;
   late Future<void> _future;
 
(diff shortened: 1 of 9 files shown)


kooWZ:
Apologies for the messy diff -- some functions were moved.

lib/page/subpage_timetable.dart
```diff
@@ -614,11 +614,14 @@ class SemesterSelectionButtonState extends State<SemesterSelectionButton> {
     String? chosenSemester = SettingsProvider.getInstance().timetableSemester;
     if (chosenSemester == null || chosenSemester.isEmpty) {
       chosenSemester = _semesterInfo!.defaultSemesterId;
+      SettingsProvider.getInstance().timetableSemester = chosenSemester;
     }
     _selectionInfo = _semesterInfo!.semesters
         .firstWhere((element) => element.semesterId == chosenSemester!);
```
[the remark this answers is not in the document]
kooWZ: Might be good advice, but I don’t think that will happen. 
If it does, it probably means xxb is switching to a new API again.

lib/repository/fdu/time_table_repository.dart
```diff
@@ -162,76 +158,18 @@ class TimeTableRepository extends BaseRepositoryWithDio {
...
-    //       }
-    //       break;
-    //     }
-    //   }
-    // }
-    return timetable;
+  String _parseDefaultSemesterId(String semesterHtml) {
+    BeautifulSoup soup = BeautifulSoup(semesterHtml);
+    final firstOption = soup.find('option', selector: '#allSemesters option');
+    return firstOption!['value']!;
```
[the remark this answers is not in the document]
kooWZ: Same as above

lib/repository/fdu/time_table_repository.dart
```diff
@@ -16,111 +16,109 @@
...
-
-  Future<void> loginJWGL(PersonInfo info) async {
-    if (loginSession != null) {
-      try {
-        await loginSession;
-      } on DioException {
-        loginSession = _authenticateJWGL(info);
-        await loginSession!;
-      }
+  // Load TimeTable. [SettingsProvider] determines which semester to load.
```
w568w: Use `///` instead of `//` to explain entities including functions, classes, etc.

lib/repository/fdu/time_table_repository.dart
```diff
@@ -162,76 +158,18 @@ class TimeTableRepository extends BaseRepositoryWithDio {
...
-    //         course.availableWeeks![i] = course.availableWeeks![i] - 1;
-    //       }
-    //       break;
-    //     }
-    //   }
-    // }
-    return timetable;
+  String _parseDefaultSemesterId(String semesterHtml) {
+    BeautifulSoup soup = BeautifulSoup(semesterHtml);
+    final firstOption = soup.find('option', selector: '#allSemesters option');
```
w568w: When a `selector` is given, the first argument will be ignored. You can safely set it to `''`.

lib/repository/fdu/time_table_repository.dart
```diff
@@ -16,111 +16,109 @@
...
-  Future<void> loginJWGL(PersonInfo info) async {
-    if (loginSession != null) {
-      try {
-        await loginSession;
-      } on DioException {
-        loginSession = _authenticateJWGL(info);
-        await loginSession!;
-      }
+  // Load TimeTable. [SettingsProvider] determines which semester to load.
+  Future<TimeTable?> loadTimeTable(PersonInfo info,
```
w568w: I track down the path but don't see where the `info` is used. Could it be removed?

lib/repository/fdu/time_table_repository.dart
w568w, line 1: Just my thoughts: things like `SemesterInfo`, `CurrentSemesterInfo`, `TimeTableSemesterInfo` and `TimeTableExtra` keep whirling in my head. I think **their names are very unintuitive**. 

Maybe we can sort out a better way to name them…

w568w (changes requested):
Some suggestions to tackle with but not all of them, since I may review it again tomorrow. Great work BTW.

lib/repository/fdu/time_table_repository.dart
```diff
@@ -16,111 +16,109 @@
...
-      throw "Retrieval failed";
+  // Load TimeTable from FDU server.
+  Future<TimeTable?> _loadTimeTableRemotely(PersonInfo info,
+      {DateTime? startTime}) async {
+    // Determine which semester we need to load.
+    // If not stored in [SettingsProvider], we use default semester.
+    Future<CurrentSemesterInfo> getAppropriateSemesterInfo(
+        PersonInfo info) async {
+      String? semesterId = SettingsProvider.getInstance().timetableSemester;
+      String? startDate = SettingsProvider.getInstance().thisSemesterStartDate;
```
w568w: I believe the repository would better not interact with `SettingsProvider` directly. 

Can you make them function arguments?

lib/repository/fdu/time_table_repository.dart
```diff
@@ -162,76 +158,18 @@ class TimeTableRepository extends BaseRepositoryWithDio {
...
-      return Cache.get<TimeTable>(
-          KEY_TIMETABLE_CACHE,
-          () async =>
-              (await loadTimeTableRemotely(info, startTime: startTime))!,
-          (cachedValue) => TimeTable.fromJson(jsonDecode(cachedValue!)),
-          (object) => jsonEncode(object.toJson()));
-    }
+  CurrentSemesterInfo _parseDefaultSemesterInfo(String semesterHtml) {
+    String defaultSemesterId = _parseDefaultSemesterId(semesterHtml);
+    TimeTableSemesterInfo semesters = _parseSemesters(semesterHtml);
```
w568w: I didn't get it. Why does `_parseDefaultSemesterInfo` call `_parseDefaultSemesterId` twice (direct & indirect via `_parseSemesters`)?

PR #575, opened 2025-09-02
feat: introduces the ability to save and manage used proxies

sseu-buhzzi:
Key changes include:

- Saved Proxies: Users can now save multiple proxy configurations.
- Proxy Management:
  - A new menu allows users to select from their saved proxies.
  - Options to "Do not use proxies", "Add a new proxy", and "Remove a proxy".
  - The previous prompts to input the proxy is now moved to the "Add a new proxy" part.
  - Removing a proxy allows selecting from the saved list to delete.
- SettingsProvider Update: A new key `KEY_SAVED_PROXIES` is added to store the list of proxies in preferences.
- Localization: New strings for proxy management actions have been added for English, Japanese, and Simplified Chinese.

Previously, users could only set a single proxy by manually inputting the address each time. This change provides a more convenient way to switch between or disable proxies.


w568w:
Thanks for the PR! I'll review it ASAP.

**I've requested an automated review from Copilot, but you needn't respond to its AI-generated suggestions. It's for quick summary only.**

lib/provider/settings_provider.dart
```diff
@@ -144,6 +145,13 @@ class SettingsProvider with ChangeNotifier {
     notifyListeners();
   }
 
+  List<String> get savedProxies => preferences!.getStringList(KEY_SAVED_PROXIES) ?? [];
```
w568w: You'd better use `List<String>.empty()` instead of `[]`. 

The former creates an ungrowable list while the latter creates a growable one. We prefer an ungrowable one, which prevents unexpected direct modification to the list (e.g. `provider.savedProxies.add("xxx")`, which has no effects on the stored value at all).

lib/page/subpage_settings.dart
```diff
@@ -268,6 +268,86 @@ class SettingsPageState extends State<SettingsPage> {
     return list;
   }
 
+  List<Widget> _buildProxyList(BuildContext menuContext) {
+    onTapListener(String? proxyUrl) {
+      SettingsProvider.getInstance().proxy = proxyUrl;
+      setState(() {});
```
w568w: Two thoughts:

1. You don't need the `setState()` here. `SettingsProvider` *is* a provider and the `context.select<SettingsProvider, String?>` below is listening to it. If `proxy` changes, the page will refresh itself automatically.
2. Rather than `SettingsProvider.getInstance()`, you'd better use `menuContext.read<SettingsProvider>()` for code uniformity. Same for the below `getInstance` calls.

lib/page/subpage_settings.dart
```diff
@@ -268,6 +268,86 @@ class SettingsPageState extends State<SettingsPage> {
...
+              S.of(context).proxy_setting_input_title,
+              initialText: context.read<SettingsProvider>().proxy,
+              hintText: S.of(context).proxy_setting_input_hint);
+          if (!context.mounted || addr == null) {
+            return; // return if cancelled
+          }
+          if (addr.isEmpty) {
+            addr = null;
+          } else {
+            var proxies = SettingsProvider.getInstance().savedProxies;
```
w568w: `final` is preferable over `var` if the variable is not changed. (And it *isn't* here, indeed.)

lib/page/subpage_settings.dart
```diff
@@ -268,6 +268,86 @@ class SettingsPageState extends State<SettingsPage> {
...
+
+  List<Widget> _buildRemoveProxyList(BuildContext menuContext) {
+    onTapListener(String proxyUrl) {
+      var proxies = SettingsProvider.getInstance().savedProxies;
+      proxies.remove(proxyUrl);
+      SettingsProvider.getInstance().savedProxies = proxies;
+      if (SettingsProvider.getInstance().proxy == proxyUrl) {
+        SettingsProvider.getInstance().proxy = null;
+      }
+      setState(() {});
```
w568w: Ditto.

lib/page/subpage_settings.dart
```diff
@@ -268,6 +268,86 @@ class SettingsPageState extends State<SettingsPage> {
...
+                PlatformContextMenu(
+                    actions: _buildRemoveProxyList(sheetContext),
+                    cancelButton: CupertinoActionSheetAction(
+                        child: Text(S.of(sheetContext).cancel),
+                        onPressed: () =>
+                            Navigator.of(sheetContext).pop()))),
+      ),
+    ];
+
+    for (var proxyUrl in SettingsProvider.getInstance().savedProxies) {
```
w568w: `var` -> `final`?

w568w (changes requested):
Mostly LGTM, minor changes required.

lib/page/subpage_settings.dart
```diff
@@ -268,6 +268,89 @@ class SettingsPageState extends State<SettingsPage> {
...
+          }
+          if (addr.isEmpty) {
+            addr = null;
+          } else {
+            final proxies = List<String>.from(context.read<SettingsProvider>().savedProxies);
+            if (proxies.contains(addr)) {
+              Noticing.showNotice(
+                context,
+                S.of(context).proxy_setting_already_exists,
+              );
```
sseu-buhzzi, line 301: Should we `await` here? Or it is fine to call without awaiting?

w568w: > Should we `await` here? Or it is fine to call without awaiting?

I believe you don't need to `await`.

lib/page/subpage_settings.dart
sseu-buhzzi: I noticed that `SettingsProvider.getInstance()` is used in `_deleteAllDataAndExit`, `_buildCampusAreaList`, `_buildLanguageList`, and `_buildThemeList`.  
For consistency, I followed the same approach in `_buildProxyList` and `_buildRemoveProxyList` in the first commit.  
Would you recommend keeping it this way, or switching to `context.read<SettingsProvider>()` instead?

w568w: > I noticed that `SettingsProvider.getInstance()` is used in `_deleteAllDataAndExit`, `_buildCampusAreaList`, `_buildLanguageList`, and `_buildThemeList`. For consistency, I followed the same approach in `_buildProxyList` and `_buildRemoveProxyList` in the first commit. Would you recommend keeping it this way, or switching to `context.read<SettingsProvider>()` instead?

Yeah… We have some inconsistency back in the old commits, but we plan to refactor them someday (since there are only syntax differences and they don't affect the app's behavior, however, we deprioritize them for some time), so we don't accept them in newly created PRs.

w568w:
Note: we're working on the `refactor-network` branch actively atm (check out the pinned issue #542), so I'm merging the PR into that branch. When we finish the job on #542, we'll return to the `main` branch.

PR #589, opened 2025-09-08
Fix pe

w568w:
Related to #542.

**PE Feature and Repository Refactor:**
* The `FudanPERepository` now uses the new `neo_login_tool` for authentication.

**Repository and Session Management Improvements:**
* Updated cookie management in `neo_login_tool.dart` to use `TolerantCookieManager` instead of the default `CookieManager`.


w568w:
I want to merge this PR directly because it's based on koowz's contribution, which I've already reviewed. 

I've only made some necessary improvements, which are all contained within `FudanPERepository`, so in theory, there shouldn't be a significant impact.

PR #602, opened 2025-09-17
feat: impl url placeholders and replace logic of banner

AInfinity-LilacDream:
as above. initially impl a {danta_access_token} placeholder, which will be replaced with user's danta token when clicked.

### DIFF
--- a/lib/widget/forum/auto_banner.dart
+++ b/lib/widget/forum/auto_banner.dart
@@ -52,10 +52,9 @@ class AutoBannerState extends State<AutoBanner> {
   void Function(bool)? onExpand;
   List<BannerExtra?>? list;
 
-  Map<String, String> placeholderList = {
-    "{danta_access_token}":
-        SettingsProvider.getInstance().forumToken?.access ?? ""
-  };
+  static final Map<String, String Function()> placeholderList = Map.unmodifiable({
+    "{danta_access_token}": () => SettingsProvider.getInstance().forumToken?.access ?? ""
+  });
 
   @override
   void initState() {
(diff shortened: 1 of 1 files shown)


lib/widget/forum/auto_banner.dart
```diff
@@ -52,6 +52,11 @@ class AutoBannerState extends State<AutoBanner> {
   void Function(bool)? onExpand;
   List<BannerExtra?>? list;
 
+  Map<String, String> placeholderList = {
+    "{danta_access_token}":
+        SettingsProvider.getInstance().forumToken?.access ?? ""
+  };
```
w568w: 1. Make it a builder pattern to allow build with the latest token (i.e., `Map<String, String Function()>`).
2. Use `static final ... = Map.unmodifiable({...})` to prevent it from being modified.

lib/widget/forum/auto_banner.dart
```diff
@@ -64,16 +69,32 @@ class AutoBannerState extends State<AutoBanner> {
...
+    if (action.isEmpty || placeholderList.isEmpty) {
+      return action;
+    }
+
+    final pattern = placeholderList.keys.map(RegExp.escape).join('|');
+    final regex = RegExp(pattern);
+
+    return action.replaceAllMapped(regex, (match) {
+      final placeholder = match.group(0)!;
+      return placeholderList[placeholder] ?? placeholder;
```
w568w: You are replacing inside an URL, so escape the value with URL Encoding as well.

w568w (changes requested):

lib/widget/forum/auto_banner.dart
AInfinity-LilacDream: added.

lib/widget/forum/auto_banner.dart
AInfinity-LilacDream: fixed.

w568w (approved):
LGTM.

PR #609, opened 2026-01-01
fix: correct `OTFloor.copyWith` field assignment

sseu-buhzzi:
Previously, `copyWith` assigns `modified` to `like`, which seems to be nonsense. For the usage of `copyWith` in showing the history modifications, `modified` should be preserved.


w568w:
Must be typo. Ref: c25fef0a5886df06dae53d633d7ec701513dc7a6

PR #611, opened 2026-01-11
feat: adjust the UI in the Exams and Grades page

### LINKED ISSUE #617: [Feature Request] 考试与成绩页面的一些调整
**你的功能需求和某个bug有关吗？**
否

**你想要什么样的功能？**
考试与成绩页面：

  - 切换学期时无需重新加载 GPA 排名，这和学期无关。
  - 加入下拉刷新。
  - GPA 排名可以选择全院系／本专业。也可以考虑让用户选择 GPA 排名内出现过的某个专业，不局限于自身专业。
  - 加入排名百分位，按院系或专业。

sseu-buhzzi:
Closes #617.

Main changes:

  - Keep GPA list alive across different semesters. When switching semesters, the already fetched GPA list is kept, and only when reentering or refreshing the page, the GPA list is fetched again.
  - Add pull-to-refresh functionality to the page.
  - Add a toggle button to switch the GPA list between all students in the department and those in the same major.
  - Add a GPA percentile column.

### DIFF
--- a/lib/page/dashboard/exam_detail.dart
+++ b/lib/page/dashboard/exam_detail.dart
@@ -16,6 +16,7 @@
  */
 
 import 'dart:io';
+import 'dart:math';
 
 import 'package:collection/collection.dart';
 import 'package:dan_xi/generated/l10n.dart';
@@ -184,9 +185,12 @@ class ExamList extends HookConsumerWidget {
       List<SemesterInfo> semesters, ValueNotifier<int?> currentSemesterIndex,
       {ValueNotifier<List<Exam>?>? currentExamRef}) {
     final currentSemesterIndexValue =
-        currentSemesterIndex.value ?? semesters.length - 3;
+        currentSemesterIndex.value ?? max(semesters.length - 2, 0);
     final currentSemester = semesters[currentSemesterIndexValue];
-    final semesterId = currentSemester.semesterId!;
+    final semesterId = currentSemester.semesterId;
+    if (semesterId == null) {
+      return _getNoDataWidget(context);
+    }
 
     final currentExamProvider = examProvider(semesterId);
     final currentScoreProvider = examScoreProvider(semesterId);
@@ -225,11 +229,7 @@ class ExamList extends HookConsumerWidget {
             AsyncError(error: final scoreError)
           )
           when examError is SemesterNoExamException && scoreError is RangeError:
-        body = Padding(
-            padding: const EdgeInsets.symmetric(horizontal: 32),
-            child: Center(
-              child: Text(S.of(context).no_data),
-            ));
+        body = _getNoDataWidget(context);
       // Other cases, such as loading or error
       case (AsyncError(:final error, :final stackTrace), _):
         body = ErrorPageWidget.buildWidget(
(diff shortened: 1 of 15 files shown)


w568w:
Our Contribution Rules:

* All substantive changes require a PR (trivial changes like formatting or adding missed files are excluded).
* **If the PR links to an Issue:**
  1. Use `Closes/Fixes #issue_number.` in the PR body to link with the issue.
  2. No need for Labels on the PR.
  3. Add the **Issue** (if haven't) to the target Milestone.
* **If the PR is independent (no Issue):**
  1. Add Labels to the PR.
  2. Add the **PR** itself o the target Milestone.

lib/page/dashboard/gpa_table.dart
w568w, line 1: You might have noticed that the project currently uses a mix of Flutter `setState`, Provider, and Riverpod & Hooks. This is a result of our ongoing, long-term refactoring process.

Our current guideline is that if you're making substantial changes to a specific UI, we'd encourage you to take the opportunity to migrate it to Riverpod. 

To be clear, this is NOT a mandatory requirement for merging, but it is recommended.

lib/page/dashboard/exam_detail.dart
```diff
@@ -521,4 +588,20 @@ class ExamList extends HookConsumerWidget {
...
+    ref.invalidate(semesterProvider);
+
+    if (semesterId != null) {
+      ref.invalidate(examProvider(semesterId));
+      ref.invalidate(examScoreProvider(semesterId));
+    }
+
+    if (isFallback) {
+      ref.invalidate(examScoreFromDataCenterProvider);
+    }
```
w568w: Calling `invalidate` here is perfectly safe. It simply discards the current value; it won't immediately trigger a UI rebuild or a network refresh if the provider isn't currently being watched.

lib/page/dashboard/exam_detail.dart
```diff
@@ -42,9 +42,55 @@ import 'package:share_plus/share_plus.dart';
 
 part 'exam_detail.g.dart';
 
+// To preserve GPA data across semesters switches, we need `keepAlive: true`.
+// To control the behaviour of dropping previous GPA data when quit and re-enter
+// the `ExamList`, we need a notifier.
```
w568w: I’m trying to get a better sense of the logic here. **Are you creating this custom Notifier specifically** to invalidate the previous value whenever the page is re-entered?

It feels a bit redundant: first of all, letting the user see the cached data from their last visit isn't necessarily a bad thing. But if you feel this reset is essential, wouldn't it be simpler to just call `ref.invalidate(gpaProvider)` inside a `postFrameCallback`?

w568w (changes requested):
Mostly LGTM!

lib/page/dashboard/exam_detail.dart
sseu-buhzzi: It's mainly because that, using `ref.invalidate(gpaProvider)` without customizing `GpaNotifier`, the provider does not enter a loading state during pull-to-refresh, which makes the data appear unchanged to the user. Therefore, I added logic to explicitly transition it into a loading state.

w568w: > the provider does not enter a loading state during pull-to-refresh

If I don't get it wrong, it is a feature rather than a bug: <https://github.com/rrousselGit/riverpod/issues/1568>. [Workaround](https://github.com/rrousselGit/riverpod/issues/1568#issuecomment-1218041038)

sseu-buhzzi: > it is a feature rather than a bug

I see it. Then I would just keep the old riverpod functions.

lib/page/dashboard/gpa_table.dart
sseu-buhzzi: I'll do this when I'm available, maybe this weekend.

PR #612, opened 2026-01-11
Update deps

w568w:
As the commit messages say.


w568w:
Have tested it myself on Android and Linux.

PR #619, opened 2026-01-13
fix: adjust default semester index in `ExamList`

### LINKED ISSUE #618: [BUG] 考试与成绩页面出现 `RangeError (length)`
**描述 Bug**
考试与成绩页面出现 `RangeError (length): Invalid value: Not in inclusive range 0..1: -1`。

**复现步骤**
复现该 Bug 的步骤：
1. 找一个复旦大一学生
2. 进入考试与成绩
3. 有几率出现这个错误

**截图**

Image

**系统信息**
 - 旦挞校园助手版本（见「设置-关于本应用」）：Version 1.4.9 build 346 #8735e32

sseu-buhzzi:
Fixes #618.

New API returns extra one semester, then users who only get two semesters will trigger a `RangeError (length)`. The commit fixes this by doing minus `2` instead of `3` at for the fallback default index, and wrapping it with a `max`, ensuring it does not result in an invalid index.

### DIFF
--- a/lib/repository/fdu/edu_service_repository.dart
+++ b/lib/repository/fdu/edu_service_repository.dart
@@ -190,14 +190,14 @@ class EduServiceRepository extends BaseRepositoryWithDio {
           continue;
         }
 
-        var courseId = "";
-        var courseName = "";
-        var type = "";
-        var category = "";
-        var date = "";
-        var time = "";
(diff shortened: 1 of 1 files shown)


w568w (approved):
LGTM!

PR #630, opened 2026-01-18
feat: improve semester parsing and add runtime data validation

sseu-buhzzi:
- Fixed the problem of failing recognizing summer semesters.
- Introduced `require` and `requireNotNull` utility functions to facilitate runtime type validation and nullability assertions with custom error messages.
- Improved data models in `edu_service_repository.dart`.

### DIFF
--- a/lib/repository/fdu/edu_service_repository.dart
+++ b/lib/repository/fdu/edu_service_repository.dart
@@ -15,6 +15,8 @@
  *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+import 'dart:convert';
+
 import 'package:beautiful_soup_dart/beautiful_soup.dart';
 import 'package:dan_xi/generated/l10n.dart';
 import 'package:dan_xi/repository/base_repository.dart';
@@ -72,15 +74,25 @@ class EduServiceRepository extends BaseRepositoryWithDio {
       path: getSemesterCourseTableUrl(defaultSemesterId),
     );
     return FudanSession.request(options, (res) {
-      final contextGetter = () => (options, res);
-
-      final Map<String, dynamic> data = require(res.data, contextGetter);
-      final List<dynamic> vms = require(data["studentTableVms"], contextGetter);
-      final Map<String, dynamic> vm = require(vms[0], contextGetter);
+      final Map<String, dynamic> data = require(
+        res.data,
+        () => FudanApiException(options.uri.toString()),
+      );
+      final List<dynamic> vms = require(
+        data["studentTableVms"],
+        () => FudanApiException(jsonEncode(data)),
+      );
+      final Map<String, dynamic> vm = require(
+        vms[0],
+        () => FudanApiException(jsonEncode(vms)),
+      );
       // Intentionally typed as int to fail fast if backend changes the schema.
       // If it was `final int = vm["id"].toString;`, it would be hard to catch
       // the bug.
-      final int idInt = require(vm["id"], contextGetter);
+      final int idInt = require(
+        vm["id"],
+        () => FudanApiException(jsonEncode(vm)),
+      );
       final id = idInt.toString();
       return id;
     });
@@ -152,9 +164,10 @@ class EduServiceRepository extends BaseRepositoryWithDio {
       path: getExamArrangeUrl(studentId),
     );
     return await FudanSession.request(options, (res) {
-      final contextGetter = () => (options, res);
-
-      final String data = require(res.data, contextGetter);
+      final String data = require(
+        res.data,
+        () => FudanApiException(options.uri.toString()),
+      );
       final soup = BeautifulSoup(data);
       // Use this selector to filter out finished exams.
       final elements = soup.findAll(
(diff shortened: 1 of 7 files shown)


lib/util/type_requires.dart
```diff
@@ -0,0 +1,25 @@
...
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+T require<T, V>(V value, Function() message) {
```
w568w: Do you mean <code><b>dynamic</b> Function() message</code>?

lib/repository/fdu/edu_service_repository.dart
```diff
@@ -69,13 +72,15 @@ class EduServiceRepository extends BaseRepositoryWithDio {
       path: getSemesterCourseTableUrl(defaultSemesterId),
     );
     return FudanSession.request(options, (res) {
-      final Map<String, dynamic> data = res.data!;
-      final List<dynamic> vms = data["studentTableVms"];
-      final Map<String, dynamic> vm = vms[0];
+      final contextGetter = () => (options, res);
```
w568w: Request options and responses can be *really* long and verbose. Do you think they should be included in the `message` so they can be displayed directly in an `ErrorPageWidget`, considering we *do* have the feature that allows user to export network logs?

lib/repository/fdu/edu_service_repository.dart
```diff
@@ -388,19 +322,6 @@ class SemesterInfo {
     return SemesterInfo(json['id'], json['schoolYear'], json['name']);
```
w568w: Is `json['name']` a `SemesterSeason`?

lib/repository/fdu/edu_service_repository.dart
```diff
@@ -410,33 +331,54 @@ class SemesterInfo {
...
-    final nameMatch = nameRegex.firstMatch(name);
+  static final _nameRegex = RegExp(
+    "\\D*(\\d{4}-\\d{4})\\D+(\\d+|autumn|fall|秋|一|上|spring|春|二|下|summer|夏|暑|三|winter|冬|寒|四)\\D*",
+    caseSensitive: false,
+  );
+
+  static SemesterSeason? _normalizeSeason(String seasonRaw) {
+    final seasonInt = int.tryParse(seasonRaw);
+    if (seasonInt != null) {
+      return SemesterSeason.values.get(seasonInt - 1);
```
w568w: Does Dart contain the `List.get()` method? I think it should be `List.elementAtOrNull()`.

w568w (changes requested):

lib/repository/fdu/edu_service_repository.dart
w568w: So, if we don't need all the details, perhaps letting `!` throw the exception is more natural, and `require` isn't really needed here.

lib/repository/fdu/edu_service_repository.dart
sseu-buhzzi: It's this one.

https://github.com/DanXi-Dev/DanXi/blob/92b5bb593a8fc1a472504a6525b583fa35b8392e/lib/util/public_extension_methods.dart#L146-L149

But true, I think `List.elementAtOrNull()` matches the semantics better, that should be it.

lib/repository/fdu/edu_service_repository.dart
sseu-buhzzi: No, it isn't. This method is useless now, so I would delete it.

lib/repository/fdu/edu_service_repository.dart
w568w: > But true, I think `List.elementAtOrNull()` matches the semantics better, that should be it.

Oh, this must be legacy code from before Dart supported those utilities. Since they exist now, it's probably best to just delete this.

lib/repository/fdu/edu_service_repository.dart
sseu-buhzzi: I added `TextOverFlow.ellipsis` and `selectable: true`.

https://github.com/DanXi-Dev/DanXi/blob/92b5bb593a8fc1a472504a6525b583fa35b8392e/lib/widget/libraries/error_page_widget.dart#L144-L149
https://github.com/DanXi-Dev/DanXi/blob/92b5bb593a8fc1a472504a6525b583fa35b8392e/lib/widget/libraries/error_page_widget.dart#L162-L167

So there will be at most one line in the long and verbose message, and users can see and copy the fuller context for debugging.

w568w: > So there will be at most one line in the long and verbose message

[A few errors](https://github.com/DanXi-Dev/DanXi/commit/f5977be95afc01299c12e2e71d1729188ddfeea1) are meant to be multi-line, so I'm worried this might mess up the display. Plus, just setting `overflow` likely won't force it into one line.

I feel like we're using the wrong field here. Rather than dumping everything into **`message`** and restricting the UI, you should customize `generateUserFriendlyDescription` to parse the error, and use `generateErrorDetails` to display the specifics in **`details`**. 

Check out the commit referred above, https://github.com/DanXi-Dev/DanXi/commit/f5977be95afc01299c12e2e71d1729188ddfeea1 for reference.

PR #648, opened 2026-02-19
fix: throw `CredentialsInvalidException` when credentials are invalid

### LINKED ISSUE #647: [BUG] 登入时帐户密码错误会进入 InAppBrowser，但会因为无 PersonInfo 而无法登入
**复现步骤**
复现该 Bug 的步骤：
1. 输入错误的帐户或密码。
2. 在浏览器窗口中登入。
3. 关闭浏览器窗口后，应用中不会登入。

**系统信息**
 - 旦挞校园助手版本（见「设置-关于本应用」）：1.5.0+347

**原因**
- 此处只要无 `loginToken` 就导向增强认证。https://github.com/DanXi-Dev/DanXi/blob/157203c25a994374981f170b6b59e23f3c74373c/lib/repository/fdu/neo_login_tool.dart#L488-L491
- 但获取 cookie 需要已有的 `PersonInfo`。https://github.com/DanXi-Dev/DanXi/blob/157203c25a994374981f170b6b59e23f3c74373c/lib/util/browser_util.dart#L315-L316

sseu-buhzzi:
This change avoids the `AuthenticationInAppBrowser` being started when user credentials are invalid.

Fixes #647.

This can also be fixed by removing the `PersonInfo` requirement when extracting cookies, and adding a DOM listener to extract user credentials from `InAppBrowser` and set them with cookies.

### DIFF
--- a/lib/page/home_page.dart
+++ b/lib/page/home_page.dart
@@ -572,7 +572,9 @@ class HomePageState extends State<HomePage> with WidgetsBindingObserver {
     _credentialsInvalidSubscription.bindOnlyInvalid(
         Constant.eventBus
             .on()
-            .where((e) => e is uis.CredentialsInvalidException)
+            .where((e) =>
+                e is uis.CredentialsInvalidException ||
+                e is neo.CredentialsInvalidException)
             .listen((_) => _dealWithCredentialsInvalidException()),
         hashCode);
     _enhancedAuthSubscription.bindOnlyInvalid(
--- a/lib/repository/fdu/neo_login_tool.dart
+++ b/lib/repository/fdu/neo_login_tool.dart
@@ -5,7 +5,6 @@ import 'package:beautiful_soup_dart/beautiful_soup.dart';
 import 'package:dan_xi/model/person.dart';
 import 'package:dan_xi/provider/state_provider.dart';
 import 'package:dan_xi/repository/cookie/persistent_cookie_jar.dart';
-import 'package:dan_xi/repository/fdu/uis_login_tool.dart' as uis;
 import 'package:dan_xi/util/condition_variable.dart';
 import 'package:dan_xi/util/io/cookie_manager_fix.dart';
 import 'package:dan_xi/util/io/dio_utils.dart';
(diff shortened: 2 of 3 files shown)


lib/common/constant.dart
```diff
@@ -60,9 +60,16 @@ class Constant {
...
+  static String get defaultUserAgent => PlatformX.isMobile
+      ? DEFAULT_USER_AGENT_MOBILE
+      : DEFAULT_USER_AGENT_DESKTOP;
+
+  static const String DEFAULT_USER_AGENT_DESKTOP =
       "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.51 Safari/537.36";
 
+  static const String DEFAULT_USER_AGENT_MOBILE =
+      "Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/145.0.0.0 Mobile Safari/537.36";
+
```
w568w, line 72: Great fix! 

I’d been wondering why it was rendered as desktop view on phones. Now it makes sense that UIS actually relies on the User Agent instead of `media-query` for the layout. 

Such a classic 'Chinese dev' move… 👎

lib/page/home_page.dart
```diff
@@ -572,9 +572,7 @@ class HomePageState extends State<HomePage> with WidgetsBindingObserver {
     _credentialsInvalidSubscription.bindOnlyInvalid(
         Constant.eventBus
             .on()
-            .where((e) =>
-                e is uis.CredentialsInvalidException ||
-                e is neo.CredentialsInvalidException)
+            .where((e) => e is uis.CredentialsInvalidException)
```
w568w: I’m a bit confused about the changes here. Why was the condition for `neo.CredentialsInvalidException` removed, and why make it inherit from `uis.CredentialsInvalidException`?

sseu-buhzzi: I think this decouples the exception handling from specific authentication backends. Since the error handling logic in `login_dialog.dart` is semantically the same for both backends, using inheritance could avoid redundant checks for each  `CredentialsInvalidException` like https://github.com/DanXi-Dev/DanXi/blob/d735112199cf23b3b16dfdc30f89eb9312a9bb19/lib/widget/dialogs/login_dialog.dart#L197-L199 and https://github.com/DanXi-Dev/DanXi/blob/d735112199cf23b3b16dfdc30f89eb9312a9bb19/lib/widget/libraries/error_page_widget.dart#L98-L99.

w568w: > the error handling logic in login_dialog.dart is semantically the same for both backends

To me, it makes more sense to keep the different error states separate. The frontend's (like `home_page`, `login_dialog`, and `error_page_widget`) error-handling logic should really follow the backend's (`neo_login_tool`) design, rather than forcing the backend to shape its errors just to suit the frontend.

Think of it this way: if we have five distinct error cases—A, B, C, D, and E—and the frontend handles A, B, and C the exact same way, and handles D and E in another same way, we should still explicitly define all five. The frontend can just do `if (err is A || err is B || err is C)` and so on. We shouldn't just boil it down to two error types and make the backend throw what the frontend wants to see. That gets confusing and is honestly just wrong coupling.

Basically, the side throwing the error has all the context, so it should provide more granular error types. It’s really no big deal for the receiver to catch a few different errors, so the design should definitely be adapted to the sender, not receiver.

What do you think?

w568w (approved):

If you feel ready to merge, switch the status from "Draft" to "Ready for review".

PR #649, opened 2026-02-19
feat: implement dynamic WebVPN URL encoding

sseu-buhzzi:
It implements the equivalent JS methods in the WebVPN webpage.

### DIFF
--- a/lib/util/webvpn_proxy.dart
+++ b/lib/util/webvpn_proxy.dart
@@ -26,6 +26,15 @@ class WebvpnRequestException implements Exception {
 /// # Order of interceptors
 /// This interceptor should be added BEFORE [RedirectInterceptor] but AFTER any other interceptors.
 class WebVPNInterceptor extends Interceptor {
+  static const _allowedHosts = {
+    "www.fduhole.com",
+    "auth.fduhole.com",
+    "danke.fduhole.com",
+    "forum.fduhole.com",
+    "image.fduhole.com",
+    "yjsxk.fudan.edu.cn",
+    "10.64.130.6",
+  };
   static final _hostCache = <String, String>{};
   static const String DIRECT_CONNECT_TEST_URL = "https://forum.fduhole.com";
   static const String EXTRA_ROUTE_TYPE = "webvpn_route_type";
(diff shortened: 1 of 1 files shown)


lib/util/webvpn_proxy.dart
```diff
@@ -25,58 +26,81 @@ class WebvpnRequestException implements Exception {
...
+    return ivHex + encryptedHex.substring(0, 2 * originalLength);
+  }
+
+  static String _padText(String text, {String encodingName = "utf8"}) {
+    final blockSize = encodingName == "utf8" ? 16 : 32;
+    // Caveats: the length of String might not be the length of UTF-8 bytes. But
+    // the JS method in WebVPN is indeed written this way.
+    final remainder = text.length % blockSize;
+    return remainder == 0 ? text : text + "0" * (blockSize - remainder);
   }
```
w568w: Since it's only used in `_encrypt`, consider moving it into that method as a nested function.

lib/util/webvpn_proxy.dart
```diff
@@ -25,58 +26,81 @@ class WebvpnRequestException implements Exception {
...
   static final Uri WEBVPN_LOGIN_URL =
       Uri.parse("https://$WEBVPN_HOST/login?cas_login=true");
   static const String WEBVPN_REDIRECTED_TO_LOGIN_PREFIX = "https://$WEBVPN_HOST/login";
+  static const String _WRDVPN_VPN_IS_THE_BEST = "wrdvpnisthebest!";
 
   static Future<bool>? tryDirectSession;
   static bool? _canConnectDirectly;
 
   /// Try to translate a URL to its WebVPN equivalent. Return `null` if the URL is not supported by WebVPN.
   static String? tryTranslateUrlToWebVPN(String url) {
```
w568w, line 52: We should probably keep a domain allowlist to maintain the original logic. 

Right now, this change routes *all* `WebVPNIntercepter` traffic through the WebVPN when off-campus, but originally, only explicitly whitelisted domains were supposed to be proxied. Also, some domains might only work outside the VPN (though no examples come to my mind right now).

w568w (changes requested):

PR #655, opened 2026-02-28
feat: add homepage

### LINKED ISSUE #563: [Feature Request] 加一个大板块，集成所有包括了圆桌、茶楼、评教的帖子
**你的功能需求和某个bug有关吗？**
否

**你想要什么样的功能？**
加一个大板块，集成所有包括了圆桌、茶楼、评教的帖子

---

已经在后端实现 https://github.com/OpenTreeHole/treehole_next/pull/189/

route: /holes/_homepage
可以参考 [request/response body](https://github.com/OpenTreeHole/treehole_next/pull/189/files#diff-5b86a7d1c0ef2128711883b00753503576056013dcbb5af05de567033777aeb4R1192-R1255)，之后会更新到apifox文档中

HydrogenC:
Closes #563. Roughly tested that it work, but not thoroughly. 

Some changes are due to a file-wise reformat. It seems to adopt new formatting rules so that a lot of lines get changed.

### DIFF
--- a/lib/model/forum/division.dart
+++ b/lib/model/forum/division.dart
@@ -45,4 +45,12 @@ class OTDivision {
 
   @override
   int get hashCode => division_id!;
+
+  // A special division id `-1` is used to represent the home page.
+  // This is not the most elegant solution, but the simplest. 
+  static const int HOME_PAGE_DIVISION_ID = 9999;
+
+  // Since the name and description has to be localized,
+  // we cannot specify these strings here. 
+  static OTDivision get home_page_division => OTDivision(HOME_PAGE_DIVISION_ID, null, null, null);
 }
--- a/lib/model/forum/division.dart
+++ b/lib/model/forum/division.dart
@@ -46,7 +46,7 @@ class OTDivision {
   @override
   int get hashCode => division_id!;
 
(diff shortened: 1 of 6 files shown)


w568w:
Thanks for your contribution. The reason I haven't commented on this PR until now is that I wasn't quite sure how to design this myself without introducing awkward abstraction leaks (like the `HOME_PAGE_DIVISION_ID=9999` defined here, along with the special value handling scattered throughout the code).

I think a better approach might be to refactor the `forum_subpage` page to remove the assumption in the existing code that "there must always be a *current division*", and update all the code that relies on this assumption. One way to do this would be to represent the "current" state as an enum (e.g., `enum CurrentTab { HOMEPAGE, DIVISION(division_id) }`). I feel that using magic/special values instead of the type system to represent special cases is highly error-prone, as it's very easy for callers to miss handling certain scenarios.

What do you think?

HydrogenC:
> I feel that using magic/special values instead of the type system to represent special cases is highly error-prone, as it's very easy for callers to miss handling certain scenarios.

Is enums like this actually allowed in Dart? If I were writing Rust I suppose I would have done that.

HydrogenC:
Closed in favor of #669.

PR #657, opened 2026-03-01
chore: sync Gradle daemon JVM and toolchain resolver with Java 23

sseu-buhzzi:
Sometimes I found it leads to a build error when I switched my local JDK version. This change helps to synchronize the Gradle Daemon JVM version.

- Register Foojay toolchain resolver in `settings.gradle` for automated JDK provisioning.
- Add `gradle-daemon-jvm.properties` to specify Java 23 as the daemon JVM.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -17,6 +17,8 @@ subprojects {
                 project.plugins.hasPlugin("com.android.library")) {
             project.android.compileSdkVersion = 36
 
+            project.java.toolchain.languageVersion = JavaLanguageVersion.of(23)
+
             // override Java version to 23
             project.android.compileOptions.sourceCompatibility = JavaVersion.VERSION_23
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -17,6 +17,12 @@ subprojects {
                 project.plugins.hasPlugin("com.android.library")) {
             project.android.compileSdkVersion = 36
 
+            java {
+                toolchain {
+                    languageVersion = JavaLanguageVersion.of(23)
+                }
+            }
+
```
w568w, line 21: Maybe use `project.java.toolchain.languageVersion = ...` for uniformity?

w568w (changes requested):
Mostly LGTM.

android/build.gradle
sseu-buhzzi: OK.

Issue #142, opened 2022-02-21
Mega-Refactor (aka A head start for newcomers to contribute)


w568w:
Soon or later it will be necessary to expect for some newcomers to contribute to the project.

However, we have not done enough work to meet the community standard. I suppose it is time to make a to-do list for it.

No need to accomplish these at once, but at least rethink them when pushing new commits.

*p.s. feel free to add new items or edit them ✍️*

# Code Principle
- [ ] Clean `nested hell` in the layouts 
- [x] Write comments for complex codes (especially those in `build()`, `initState()` and repositories)

# Contributing Guide
- [x] Have a [guide](https://github.com/DanXi-Dev/DanXi/blob/main/CONTRIBUTING.md) describing how to contribute to this project
- [x] #256

w568w:
[Riverpod](https://riverpod.dev) has been chosen as our new state management library.

w568w:
We have been slowly migrating our UIs into Riverpod. Close it.

PR #668, opened 2026-03-16
feat: AI summary for forum posts

Bengerthelorf:
## Summary

Add AI-powered summary feature for forum post discussions.

- **Data model** (`ai_summary.dart`): Response models for summary data including discussion branches, user interactions, keywords, and trace ID
- **API layer** (`forum_repository.dart`): `loadAiSummary` and `submitAiSummaryFeedback` endpoints
- **Summary sheet** (`ai_summary_sheet.dart`): Bottom sheet UI with loading skeleton, polling for generation status, error handling, and feedback (upvote/downvote)
- **Entry point** (`hole_detail.dart`): Floating action button (Material / Cupertino), conditional on `ai_summary_available` field
- **Floor navigation**: Scroll-to-floor and highlight animation when tapping floor references in the summary
- **Disclaimer**: AI-generated content notice displayed at the top of the sheet
- **i18n**: Full localization for en, zh-Hans, zh-Hant, ja
- **Tests**: Unit tests for model JSON parsing


lib/model/forum/ai_summary.dart
w568w, line 1: All structs in this file should use `@JsonSerializable` to generate `from/toJson` method.

lib/model/forum/ai_summary.dart
```diff
@@ -0,0 +1,142 @@
+/*
+ *     Copyright (C) 2021  DanXi-Dev
```
w568w: 2021 -> 2026? 🤔

lib/widget/forum/ai_summary_sheet.dart
w568w, line 1: For new page, please use `riverpod` Framework. Flutter's vanilla state management is discouraged.

Refer to #142.

w568w (changes requested):
Some instant suggestions come to my mind.

lib/widget/forum/forum_widgets.dart
```diff
@@ -699,13 +695,20 @@ class OTFloorMentionWidget extends StatelessWidget {
...
-                              pagedListViewController != null) {
-                            // Scroll to the corresponding post
+                          if (inThatFloorPage) {
+                            // Scroll to the corresponding post, then highlight
                             pagedListViewController.scheduleLoadedCallback(
-                                () async => await pagedListViewController!
-                                    .scrollToItem(floor),
-                                rebuild: true);
+                                () async {
+                              await pagedListViewController
```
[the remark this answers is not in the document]
Bengerthelorf: Dart's flow analysis promotes `final` local variables even when the null check is stored in an intermediate boolean. We verified this compiles without errors — `pagedListViewController` is a `final` local, and `inThatFloorPage` is only `true` when the null check passes. Running `flutter analyze` on this file confirms no issues.

lib/widget/forum/ai_summary_sheet.dart
```diff
@@ -0,0 +1,769 @@
...
+      label: Text(
+        "#$floorNumber",
+        style: const TextStyle(fontSize: 15, fontWeight: FontWeight.w500),
+      ),
+      onPressed: () async {
+        final resolved = widget.floorResolver != null
+            ? await widget.floorResolver!(floorNumber)
+            : await ForumRepository.getInstance().loadFloorById(floorNumber);
+        if (resolved != null && context.mounted) {
+          OTFloorMentionWidget.showFloorDetail(context, resolved);
```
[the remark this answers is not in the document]
Bengerthelorf: The `floorResolver` parameter is always provided by the caller (`hole_detail.dart`), and `_resolveFloor` already handles dual matching (by `floor_id` first, then by 1-based index). The `loadFloorById` fallback is unreachable in practice.

lib/model/forum/ai_summary.dart
```diff
@@ -0,0 +1,142 @@
...
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+// Lightweight models for AI summary API responses.
+class AiSummaryResponse {
+  final int? code;
+  final String? message;
+  final AiSummaryData? data;
+
+  const AiSummaryResponse({this.code, this.message, this.data});
```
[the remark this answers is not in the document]
Bengerthelorf: The model tests were intentionally removed in commit 00e6693b.

lib/repository/forum/forum_repository.dart
```diff
@@ -484,6 +485,42 @@ class ForumRepository extends BaseRepositoryWithDio {
...
 
+  Future<AiSummaryResponse> loadAiSummary(
+    int holeId, {
+    bool? forceRefresh,
+  }) async {
+    final options = RequestOptions(
+      path: "$_BASE_URL/holes/$holeId/summary",
+      method: "GET",
+      queryParameters: {
+        if (forceRefresh != null) "force_refresh": forceRefresh,
```
sseu-buhzzi:
```suggestion
        "force_refresh": ?forceRefresh,
```

lib/widget/forum/ai_summary_sheet.dart
```diff
@@ -0,0 +1,770 @@
...
+
+    final code = response.code;
+    if (code == 1000) return response.data ?? AiSummaryData();
+    if (code != 1001 && code != 1002) {
+      throw AiSummaryApiException(code, response.message);
+    }
+    if (attempt < maxPollAttempts) {
+      await Future.delayed(pollInterval);
+    }
+  }
```
sseu-buhzzi, line 72: I think it could use a `CancelToken` like this, by adding an optional parameter `cancelToken` to `loadAiSummary`.

```suggestion
  final cancelToken = CancelToken();
  ref.onDispose(() => cancelToken.cancel("provider disposed"));

  for (int attempt = 0; attempt <= maxPollAttempts; attempt++) {
    try {
      final response = await ForumRepository.getInstance().loadAiSummary(
        holeId,
        cancelToken: cancelToken,
      );

      final code = response.code;
      if (code == 1000) return response.data ?? AiSummaryData();
      if (code != 1001 && code != 1002) {
        throw AiSummaryApiException(code, response.message);
      }
      if (attempt < maxPollAttempts) {
        await Future.delayed(pollInterval);
      }
    } on DioException catch (e) {
      if (CancelToken.isCancel(e)) {
        throw StateError('disposed');
      }
      rethrow;
    }
  }
```

Bengerthelorf: The polling wait time is mostly spent on the `Future.delayed` between requests — the HTTP requests themselves return pretty quickly. I think `CancelToken` can only cancel `Dio` requests and doesn't really have control over `Future.delayed`, so in practice it would probably end up being essentially equivalent to the current `disposed` flag approach anyway.

sseu-buhzzi: OK, then the current is fine.

sseu-buhzzi (approved):

lib/repository/forum/forum_repository.dart
w568w: Crazy grammar. Is Dart going to be next Kotlin?

sseu-buhzzi: Actually Kotlin doesn't seem that crazy.
```kotlin
buildMap {
    // ...
    forceRefresh?.let { put("force_refresh", it) }
}
```

lib/model/forum/ai_summary.dart
```diff
@@ -0,0 +1,112 @@
...
+
+import 'package:json_annotation/json_annotation.dart';
+
+part 'ai_summary.g.dart';
+
+@JsonSerializable()
+class AiSummaryResponse {
+  int? code;
+  String? message;
+  AiSummaryData? data;
```
w568w: AI must have read the original code and decide that, "every field in JSON must be nullable", which is false. We do NOT have the requirement; it's BAD practice.

In fact, we encourage the opposite behavior: fields in JSON should be non-null as much as possible. We are slowly refactoring the remaining old code towards the goal.

lib/page/forum/hole_detail.dart
```diff
@@ -489,6 +498,16 @@ class BBSPostDetailState extends State<BBSPostDetail> {
...
       backgroundColor: Theme.of(context).scaffoldBackgroundColor,
+      material: (context, platform) => MaterialScaffoldData(
+        floatingActionButton: showAiSummaryEntry
+            ? FloatingActionButton.extended(
+                onPressed: _isOpeningSummary ? null : _openAiSummarySheet,
+                icon: const Icon(Icons.auto_awesome),
+                label: Text(S.of(context).ai_summary_title),
+              )
+            : null,
+        floatingActionButtonLocation: FloatingActionButtonLocation.endFloat,
```
w568w, line 507: Will it be annoying to constantly show a floating button on the post detail page? Especially consider that our reply button is not so obvious as it, which could emphasize the wrong UX concept that "this button is for creating a new reply".

Maybe we should make it an optional feature that can be disabled in the setting.

lib/page/forum/hole_detail.dart
```diff
@@ -639,54 +658,173 @@ class BBSPostDetailState extends State<BBSPostDetail> {
...
-  // Load all floors, in case we have to scroll to end or to a specific floor.
-  Future<List<OTFloor>> _loadAllContent() async {
-    final allFloors = await ForumRepository.getInstance()
-        .loadFloors((_renderModel as Normal).hole, offset: 0, size: 0);
+  bool _shouldShowAiSummaryEntry() {
+    if (_renderModel case Normal(hole: var hole)) {
+      return hole.ai_summary_available == true;
+    }
+    return false;
+  }
```
w568w: It'd better be a getter rather than a function.

lib/page/forum/hole_detail.dart
```diff
@@ -639,54 +658,173 @@ class BBSPostDetailState extends State<BBSPostDetail> {
...
+        target = allFloors[index];
+      }
     }
+    return target;
+  }
 
-    _listViewController.replaceAllDataWith(allFloors);
-    _allDataLoaded = true;
+  Future<void> _locateFloor(OTFloor floor) async {
+    Navigator.of(context).popUntil((route) => route is! PopupRoute);
```
w568w: It feels dangerous to me. Generally, the dialog logic should be well-structured: for instance, wherever a dialog is opened, exactly one `pop` should be executed on every possible exit path (i.e., the point after the `await`). Thus other logic should NOT care about the navigator, routing, etc. at all.

Using `popUntil` here relies heavily on a specific assumption—namely, that routes which shouldn't be popped definitely do not use a `PopupRoute`. I don't think `_locateFloor` should depend on this kind of implicit logic.

lib/page/forum/hole_detail.dart
```diff
@@ -639,54 +658,173 @@ class BBSPostDetailState extends State<BBSPostDetail> {
...
+    try {
+      final allFloors = await ForumRepository.getInstance()
+          .loadFloors((_renderModel as Normal).hole, offset: 0, size: 0);
+
+      if (allFloors == null) {
+        throw Exception("Failed to fetch all floors");
+      }
+
+      _listViewController.replaceAllDataWith(allFloors);
+      _allDataLoaded = true;
```
w568w: `_allDataLoaded` is read in `_loadAllContent` but written in `_doLoadAllContent`, which seems strange. Consider moving all logic into `_loadAllContent`? (Hint: Use `.then` or make `_loadAllContent` async)

lib/widget/forum/ai_summary_sheet.dart
```diff
@@ -0,0 +1,770 @@
...
+        if (status >= 500) return S.of(context).ai_summary_server_error;
+      }
+      if (error.type == DioExceptionType.connectionTimeout ||
+          error.type == DioExceptionType.sendTimeout ||
+          error.type == DioExceptionType.receiveTimeout ||
+          error.type == DioExceptionType.connectionError) {
+        return S.of(context).ai_summary_network_error;
+      }
+    }
+    return S.of(context).operation_failed;
```
w568w: Consider falling back to `ErrorPageWidget.generateUserFriendlyDescription`.

w568w (changes requested):
Mostly LGTM. Please check my final comments here.

lib/model/forum/ai_summary.dart
Bengerthelorf: Made fields non-null with defaults where the backend guarantees them. `cache_expires_at` and `is_cached` kept nullable as they're planned but not yet implemented in the backend.

lib/page/forum/hole_detail.dart
Bengerthelorf: Added. Default on.

lib/model/forum/ai_summary.dart
```diff
@@ -23,11 +23,11 @@ part 'ai_summary.g.dart';
...
 class AiSummaryResponse {
-  int? code;
-  String? message;
-  AiSummaryData? data;
+  final int code;
+  final String message;
+  final AiSummaryData? data;
 
-  AiSummaryResponse({this.code, this.message, this.data});
+  AiSummaryResponse({this.code = 0, this.message = '', this.data});
```
w568w: AI made another mistake here: confused by the error message, it arbitrarily assigned default values. 

This is **completely wrong** and essentially defeats the purpose of making it non-null. 

The correct approach is to either use the `required` keyword or convert them into positional parameters (by removing the surrounding `{}`).

lib/page/forum/hole_detail.dart
```diff
@@ -406,7 +406,7 @@ class BBSPostDetailState extends State<BBSPostDetail> {
 
   @override
   Widget build(BuildContext context) {
-    final bool showAiSummaryEntry = _shouldShowAiSummaryEntry();
+    final bool showAiSummaryEntry = _shouldShowAiSummaryEntry;
```
w568w: Inline it.

lib/widget/forum/ai_summary_sheet.dart
Bengerthelorf: Also removed the unused `ai_summary_network_error` l10n key since those cases are now handled by `ErrorPageWidget`.

lib/page/forum/hole_detail.dart
```diff
@@ -805,26 +805,19 @@ class BBSPostDetailState extends State<BBSPostDetail> {
...
       _listViewController.replaceAllDataWith(allFloors);
       _allDataLoaded = true;
-
       return allFloors;
-    } catch (e) {
+    }).catchError((Object e) {
       _loadAllContentFuture = null;
-      rethrow;
-    }
+      throw e;
```
w568w: If you find yourself repeatedly chaining primitives like `Future.value()`, `Future.then`, and `Future.catchError`, consider simplifying the code by converting it to an `async` function and using `await`.

lib/provider/settings_provider.dart
```diff
@@ -660,6 +661,18 @@ class SettingsProvider with ChangeNotifier {
     notifyListeners();
   }
 
+  bool get isAiSummaryEnabled {
+    if (preferences!.containsKey(KEY_AI_SUMMARY_ENABLED)) {
+      return preferences!.getBool(KEY_AI_SUMMARY_ENABLED)!;
+    }
+    return true;
```
w568w, line 668: I thought it's better to be an opt-in rather than opt-out? Need discussion.

lib/page/forum/hole_detail.dart
Bengerthelorf: Changed to a single `Navigator.pop`. The floor detail modal is already popped by the caller before `onLocate`.

lib/widget/forum/ai_summary_sheet.dart
```diff
@@ -113,16 +114,10 @@ class AiSummarySheet extends ConsumerWidget {
...
-      if (error.type == DioExceptionType.connectionTimeout ||
-          error.type == DioExceptionType.sendTimeout ||
-          error.type == DioExceptionType.receiveTimeout ||
-          error.type == DioExceptionType.connectionError) {
-        return S.of(context).ai_summary_network_error;
       }
     }
-    return S.of(context).operation_failed;
+    return ErrorPageWidget.generateUserFriendlyDescription(
+        S.of(context), error);
```
w568w: You'd better carry the stack trace with it. Error object itself usually doesn't hold enough information for debugging.

w568w (changes requested):

lib/provider/settings_provider.dart
Bengerthelorf: Defaulting to enabled for now since we're in internal testing — participants should see the feature without having to discover the setting. Can change to opt-in after public release.

Bengerthelorf: Also planning to add multiple entry styles (e.g. a bottom bar similar to Feishu) as options.

lib/page/forum/hole_detail.dart
```diff
@@ -796,28 +794,30 @@ class BBSPostDetailState extends State<BBSPostDetail> {
...
-      _allDataLoaded = true;
-      return allFloors;
-    }).catchError((Object e) {
+    try {
+      return await (_loadAllContentFuture ??= _fetchAllContent());
+    } catch (_) {
       _loadAllContentFuture = null;
-      throw e;
-    });
+      rethrow;
```
w568w, line 808: I don't get what `try` + `catch` + `rethrow` here is doing. Isn't it equivalent to no use of try-catch?

lib/provider/settings_provider.dart
w568w: > Defaulting to enabled for now since we're in internal testing — participants should see the feature without having to discover the setting. Can change to opt-in after public release.

Make sense.

w568w (changes requested):

lib/page/forum/hole_detail.dart
Bengerthelorf: The try/catch/rethrow isn't a no-op — it clears `_loadAllContentFuture = null` before rethrowing, so the next call can retry.

Bengerthelorf: Keeping both the cache and its cleanup in the same method — moving the try/catch into `_fetchAllContent` would scatter the caching logic across two methods again.

w568w: > The try/catch/rethrow isn't a no-op — it clears `_loadAllContentFuture = null` before rethrowing, so the next call can retry.

Oh I missed the previous line. Then it makes sense.

w568w (approved):

PR #676, opened 2026-05-08
build: add a nix flake to build and run DanXi on NixOS

sseu-buhzzi:


w568w:
Sorry for the delay. I would review it tomorrow asap.

default.nix
```diff
@@ -0,0 +1,95 @@
...
+        --replace-fail
+        'URL https://github.com/nlohmann/json/releases/download/v''${NLOHMANN_JSON_VERSION}/json.tar.xz'
+        'SOURCE_DIR ${nlohmannJsonSrc}'
+      )
+      substituteInPlace "''${args[@]}"
+    '';
+in
+flutter.buildFlutterApplication rec {
+  pname = "dan_xi";
+  version = "1.5.2+349";
```
w568w: I wonder why we need to declare the version multiple times? Can Nix read the version string from the `pubspec.yaml` - which should be the single-source truth? Like what `pkgver()` function in Arch's `PKGBUILD` does.

default.nix
w568w, line 1: Can we move `nix/*`, `*.nix`, `flake.lock` to `packaging/` subdir? It might be better to keep the project root less complicated, given that it has been already very bloaty now.

nix/util/hex-to-base36.nix
w568w, line 1: Where do these two util files get invoked? I somehow failed to find them.

w568w (changes requested):

w568w:
Please also check GitHub Copilot's suggestions. Most of them look valid to me.

nix/util/hex-to-base36.nix
sseu-buhzzi: These are used to resolve some gradle paths which are still WIP, and in no use now. I forgot to remove them when switching branches.

packaging/nix/flake.nix
```diff
@@ -0,0 +1,175 @@
...
+            args=(
+              -i
+              '/^\s*android\.aapt2FromMavenOverride\s*=/d'
+              "$gradle_prop_path"
+            )
+            sed "''${args[@]}"
+          fi
+          cat >>"$gradle_prop_path" <<-EOF
+          android.aapt2FromMavenOverride=$ANDROID_HOME/build-tools/${androidBuildToolsVersion}/aapt2
+          EOF
```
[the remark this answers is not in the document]
sseu-buhzzi: The gnused does use `\s` to match whitespace.

```shell
nix shell nixpkgs#gnused -c sed 's/\s*/-/g' <<<'1   2 3'
# -1-2-3-
```

default.nix
sseu-buhzzi: For `/nix/*` we can move them into a subdir. By the way `flake.nix` and `flake.lock` must be in the same directory.

When `flake.nix` and `flake.lock` are in a subdir such as `/packaging/nix/`, it would require `nix build github:DanXi-Dev/DanXi?dir=packaging/nix` to build.

w568w: > When `flake.nix` and `flake.lock` are in a subdir such as `/packaging/nix/`, it would require `nix build github:DanXi-Dev/DanXi?dir=packaging/nix` to build.

I think this is acceptable. From what I recall, packaging scripts like this are typically kept in a separate repo (as is the case with Arch and Gentoo) rather than sharing the same repository as the software itself. 

Since we are adopting a monorepo approach, placing it in a subdirectory might be ***semantically*** more appropriate. 

(Alternatively, we could also create an independent repository, like `DanXi-Dev/DanXi.nix`?) Let me know what you think.

sseu-buhzzi: I agree that creating a separate repository for me to maintain makes a lot of sense. Whenever the mainline updates its dependencies, these Nix packaging files need to be updated accordingly. Keeping them in the main repo would introduce too much noise, leading to unnecessary commits and PRs that are irrelevant to core development.

I will then create a [DanXi-Dev/DanXi-nix](https://github.com/DanXi-Dev/DanXi-nix) (The other repos are using `DanXi-*` too). And for this PR #676 I will scope it down to an update for READMEs.

w568w (approved):

PR #681, opened 2026-06-12
feat: add post filter for holes and floors

### LINKED ISSUE #576: [Feature Request]楼层添加对键盘上Home,Pgup,Pgdn,End等键的支持，添加搜索功能
**你的功能需求和某个bug有关吗？**
否

**你想要什么样的功能？**

在茶楼和评教的楼层添加对键盘上Home,Pgup,Pgdn,End等键的支持，或者方向键，以方便浏览。在楼里再添加搜索框或者Ctrl+F的搜索快捷键，方便在当前楼进一步查找所需内容。

### LINKED ISSUE #680: [Feature Request] 内置搜索列以筛选贴子
**你的功能需求和某个bug有关吗？**
`否`

**你想要什么样的功能？**
简要、清晰地描述你请求增加的功能。

树洞可不可以考虑在【订阅】/【收藏】这两栏里内置搜索栏，收藏的帖子多了找起来也有点麻烦还有可能找不到
如果在每个帖子内部也能有搜索栏就好了，想查某条特定的回复也不用一层层翻（特别是那些很高的楼）

sseu-buhzzi:
image

Close #680.


sseu-buhzzi:
image

One problem is that buttons are now too many in the forum page. I am not sure where this new filter button should go.

w568w:
Sorry for late reply. The idea LGTM, but maybe the search toolbar is too hard-core for average users? :thinking: 

I'm considering an "easy mode", which allows user to select filters (e.g., `Like: [✔Yes] [No] [Unspecified]`) . 

Also, you might want to support showing a dropdown filter history list in the text field, like the post searching page we have currently, so that user doesn't need to input the same filter again and again.

sseu-buhzzi:
User friendly UI selectors are like:
image
image

History and UI restoring are like:
image

sseu-buhzzi:
This PR is ready for review now.

lib/model/forum/floor.dart
```diff
@@ -41,6 +41,8 @@ class OTFloor {
   List<OTFloor>? mention;
   int? dislike;
   bool? disliked;
+  @JsonKey(includeFromJson: false, includeToJson: false)
+  Object? meta;
```
w568w: It's better to have an explicit typed field (e.g., `PostFilterPlaceholderHint? hint`) for this and comment about its usage. 

`Object meta` is more general but it goes *way too* far, I think.

lib/model/forum/hole.dart
```diff
@@ -40,6 +40,8 @@ class OTHole {
   int? favorite_count;
   int? subscription_count;
   bool? ai_summary_available;
+  @JsonKey(includeFromJson: false, includeToJson: false)
+  Object? meta;
```
w568w: Ditto.

lib/model/forum/hole.dart
```diff
@@ -68,7 +104,7 @@ class OTHole {
       time_deleted!.isNotEmpty &&
       DateTime.tryParse(time_deleted!) != null;
 
-  static final DUMMY_POST = OTHole.dummy();
+  static final dummyPost = OTHole.dummy();
```
w568w: We *did* select using UPPER_CASE in our project for global constants. 

It's against Dart's current default naming style but we did our choice even before the style is founded, so I don't think it's a big problem. 

Although, maybe in a following PR we can rename them all, not in this one.

lib/page/forum/hole_detail.dart
```diff
@@ -433,7 +455,7 @@ class BBSPostDetailState extends State<BBSPostDetail> {
     _backgroundImage = SettingsProvider.getInstance().backgroundImage;
     final pagedListView = PagedListView<OTFloor>(
       pagedController: _listViewController,
-      noneItem: OTFloor.dummy(),
+      noneItem: OTFloor.dummyPost,
```
w568w: Why use a constant here? I thought we've overridden equality `=` for the `OTFloor`?

lib/page/subpage_forum.dart
```diff
@@ -497,15 +521,25 @@ class ForumSubpageState extends PlatformSubpageState<ForumSubpage> {
...
   }
 
+  List<OTHole>? _applyPostFilter(List<OTHole> posts) {
+    if (posts.isEmpty) return null;
+    final filtered = [
+      for (final post in posts)
+        if (_postFilter.holeMatches(post)) post,
+    ];
+    return filtered.isEmpty
+        ? [posts.last.copyWith(meta: PostFilterPlaceholderHint(posts.length))]
```
w568w: I got what you want to achieve here, but there's a gotcha: `PagedListView` will add `Dismissible` for all post items, so user can swipe the hint widget to hide an invisible (but real) post.

lib/page/forum/hole_detail.dart
```diff
@@ -661,30 +692,53 @@ class BBSPostDetailState extends State<BBSPostDetail> {
...
-                      (_renderModel as Normal).selectedPerson = null;
-                    }
-                    await refreshListView();
-                  },
-                  child: pagedListView),
-              _ => pagedListView,
-            },
+            child: WithPostFilterBar(
+              filter: _postFilter,
+              onApply: (expr) => setState(() => _postFilter.apply(expr)),
```
w568w: Will `setState(() => _postFilter.apply(expr))` trigger a refresh of the list view? Otherwises, the shown items on the screen will not be refiltered.

lib/page/forum/hole_detail.dart
w568w: We use the current `_listViewController.length()` as the offset for next loading. So if you filter the post list data, maybe the offset will be wrong here.

w568w: It puts unfiltered data to the list view directly. Will it bypass an enabled filter?

lib/page/subpage_forum.dart
```diff
@@ -708,13 +753,20 @@ class ForumSubpageState extends PlatformSubpageState<ForumSubpage> {
...
           ),
           body: Builder(
             // The builder widget updates context so that MediaQuery below can use the correct context (that is, Scaffold considered)
             builder: (context) => _buildPageBody(context, false),
           ),
         );
       case PostsType.NORMAL_POSTS:
+        return _buildPageBody(context, PlatformX.isMaterial(context));
       case PostsType.EXTERNAL_VIEW:
         return _buildPageBody(context, PlatformX.isMaterial(context));
```
w568w, line 778: Looks unnecessary to me. `case xxx: case yyy: ...` will fall both cases down to the statement following it.

w568w (changes requested):

sseu-buhzzi:
Sorry for not noticing the reviews recently, I will apply the commented changes as soon as possible.

lib/page/forum/hole_detail.dart
sseu-buhzzi: OTHole always uses such a constant `OTHole.DUMMY_POST`, so I added it to keep the two models similar.

https://github.com/DanXi-Dev/DanXi/blob/934533c66e51e7db82776df5339d94105f00610d/lib/page/subpage_bbs.dart#L424-L426
https://github.com/DanXi-Dev/DanXi/blob/934533c66e51e7db82776df5339d94105f00610d/lib/page/subpage_bbs.dart#L261-L263

If `BBSPostDetailState._loadContent` ever in the future needs to return `[OTFloor.DUMMY_POST]` as the non-ending empty list, it can then be written the same way `ForumSubpageState` was.

I will update them to `DUMMY_POST` instead of `dummyPost`.

lib/page/forum/hole_detail.dart
sseu-buhzzi: No, it will not be refreshed. In the main "forum subpage" or the "hole detail" page the user can refresh to get the filtered view. 

But in some other render models like "my replies", it does not support pull-down-refresh, so it would be a bug there that the filter never filters the posts. I will fix it.

lib/page/forum/hole_detail.dart
sseu-buhzzi: https://github.com/DanXi-Dev/DanXi/blob/94f702a44026aa4238cef0847ee7358651610bf9/lib/page/forum/hole_detail.dart#L154-L159

This person filter that has already been existing could also generate wrong offsets with using `_listViewController.length()` as `startFloor`. I think the fix could be to replace some of these `.length()` occurrences with `page * Constant.POST_COUNT_PER_PAGE` like the Normal render model.

lib/page/subpage_forum.dart
sseu-buhzzi: I could not remember why I was adding it...

w568w:
Sorry for the laaaaaaate reply. I'm so busy recently, thus having no time to review it. I might get it done this week.

sseu-buhzzi:
I will try to implement the support for keyboard shortcuts like `<C-F>` and `/` next weekend. I would be busy in the coming workdays too.

w568w:
Let's make it in 1.5.4 or 1.6.0.

Version 1.5.3 should be released today since a new school year has begun.

PR #685, opened 2026-07-09
Fix 2FA cookie import & ElectricityHistory API

w568w:
This PR:

- fixes the 2FA logging in issue (in which the 2FA can never be authenticated; it always jumps back to non-login state) by changing cookie importing behavior;
- updates with the ElectricityHistory API change.


sseu-buhzzi (approved):
Sorry for the delayed review. All LGTM, however I no longer have access the campus dormitory system APIs or a graduate student account, so I was not able to actually to test or verify the format of the responses they return.

PR #688, opened 2026-08-10
fix(windows): restore Simplified Chinese font fallback

KawaiiSh1zuku:
Use the zh-CN locale for Simplified Chinese so Flutter selects the correct Windows CJK glyphs while retaining zh-Hant for Traditional Chinese.

Apply MSVC /permissive only to gal_plugin, enabling building on lower Windows SDK version.

Add locale regression tests for Simplified and Traditional Chinese.

Validated with flutter test and flutter build windows --debug.

Before:
before

After:
after

_* LLM usage claim: some code is generated with the assistance of gpt-5.6-sol_


w568w (approved):

PR #708, opened 2026-09-09
fix(ai-summary): use empty lists when the backend responds with nulls

sseu-buhzzi:
The backend can responds with:
```
{
  code: 1001,
  message: "processing",
  data: {
    hole_id: 0,
    trace_id: "",
    summary: "",
    branches: null,
    interactions: null,
    keywords: null,
    generated_at: "",
  },
}
```

Therefore default values are needed for backend-nullable fields. Or as another fix, the backend can return data as `null` or `{}`.


w568w (approved):
Woah, I somehow missed this PR. Sorry

PR #713, opened 2026-09-22
feat(forum): add multi-tag filtering for hole lists

fduTristin:
The treehole_next backend now accepts multiple tags on both GET /holes (query param "tags") and GET /holes/_homepage, returning only holes that carry ALL the given tags (AND semantics, implemented as a GROUP BY on hole_tags with HAVING COUNT(DISTINCT tag_id) = N).

How the frontend talks to the backend:
- ForumRepository.loadHoles gains a "List<String>? tags" parameter. When non-empty, the tag names are put into queryParameters["tags"] as a Dart List, so Dio encodes them as repeated query params (e.g. ?tags=a&tags=b), which fiber parses into []string.
- The parameter is wired into both branches of loadHoles: the _homepage endpoint (offset/size) and the legacy /holes endpoint (start_time/length). The legacy single "tag" param is kept for backward compatibility.

UI changes for the multi-tag filter:
- widget/forum/ottag_selector.dart: new showOTTagSelectionDialog(), a platform dialog embedding the existing OTTagSelector. Initial tag names are resolved against the full tag list (loadTags) so that OTTag equality (based on tag_id) works inside the selector. The dialog returns the selected tag names, or null when cancelled.
- page/subpage_forum.dart (forum homepage): a tag icon button is added to buildForumTopBar() next to the search button. It opens the selection dialog and pushes /bbs/discussions with the new "tagFilters" argument when at least one tag was chosen.
- page/subpage_forum.dart (ForumSubpage): FILTER_BY_TAG now stores a list of tags (_tagFilters) instead of a single _tagFilter string. It accepts both the old "tagFilter" (String) argument, used by tag chips on hole cards, and the new "tagFilters" (List<String>) argument, so existing navigation keeps working. When filtering, the request ignores the current division and passes "tags" to the backend. The filter page title shows all tags joined together, and a new trailing edit button re-opens the selection dialog to modify the active filter in place and refresh the list.
- l10n: added "filter_by_tags" and "filtering_by_tags" for en/zh_CN/zh_Hant/ja (run intl_utils:generate to regenerate S).


lib/page/subpage_forum.dart
```diff
@@ -625,10 +630,16 @@ class ForumSubpageState extends PlatformSubpageState<ForumSubpage> {
...
-      if (widget.arguments?.containsKey('tagFilter') ?? false) {
-        _tagFilter = widget.arguments!['tagFilter'];
+      if (widget.arguments?.containsKey('tagFilters') ?? false) {
+        final List<dynamic>? tagFilters = widget.arguments!['tagFilters'];
+        if (tagFilters != null) {
+          _tagFilters = tagFilters.whereType<String>().toList();
+        }
+      } else if (widget.arguments?.containsKey('tagFilter') ?? false) {
+        final String? tagFilter = widget.arguments!['tagFilter'];
+        if (tagFilter != null) _tagFilters = [tagFilter];
```
w568w, line 640: Do NOT keep backward support for `tagFilter`. It's meaningless.

w568w (changes requested):

